PR #34, opened 2019-02-15
Contributions!

emplums:
This PR adds contributions to the bottom of the page for our design guidelines!

I'd like to merge as-is, then follow up when the docs-components package is ready and replace with that component


src/Contributors.js
```diff
@@ -0,0 +1,104 @@
+import React, {useEffect, useState} from 'react'
+import PropTypes from 'prop-types'
+import {Link, Text, Avatar, Flex} from '@primer/components'
+
+function generateContributors(authors) {
+  const logins = []
+  const uniqueAuthors = authors.filter(author => {
```
emplums: There's probably a shiny-er way to do this but 🤷‍♀️

src/Contributors.js
```diff
@@ -0,0 +1,104 @@
...
+  const [authors, setAuthors] = useState([])
+  useEffect(
+    () => {
+      const url = `https://api.github.com/repos/${repoPath}/commits?path=${filePath}`
+      fetch(url)
+        .then(response => response.json())
+        .then(commits => {
+          const commitData = []
+          const ids = []
+          for (let i = 0; i < commits.length; i++) {
```
emplums: fetch the data for all the commits to the file then restructure the data and only save unique authors

src/Contributors.js
```diff
@@ -0,0 +1,104 @@
...
+                time: new Date(commits[i].commit.author.date),
+                commitUrl: commits[i].html_url
+              })
+              ids.push(commits[i].author.id)
+            }
+          }
+          setAuthors(commitData)
+        })
+    },
+    [filePath]
```
emplums: This is a Hooks thing, it means that we will only run the effect when the `filePath` prop has changed

src/Contributors.js
```diff
@@ -0,0 +1,104 @@
...
+          <Link href={`https://github.com/${lastAuthor.login}`}> {lastAuthor.login}</Link> on{' '}
+          <Link color="gray.5" href={lastAuthor.commitUrl}>{`${month} ${day}, ${year}`}</Link>
+        </Text>
+      </Flex>
+    )
+  }
+}
+
+const Contributors = ({filePath, repoPath, contributors}) => {
+  const [authors, setAuthors] = useState([])
```
emplums: set up the `authors` object in state

src/Contributors.js
shawnbot: The unique one-liner looks something like this:

```js
const unique = values.filter((value, i, list) => list.indexOf(value) === i)
```

In other words: only the values for which the index of the value is the index of the _first_ one. ✨

src/Contributors.js
shawnbot: It doesn't look like we're using the index here, so could this be a `for..of` loop?

```js
for (const commit of commits) {
  // ✨
}
```

src/Contributors.js
emplums: Sweeet thank you!

PR #43, opened 2019-04-17
don't discourage the use of kebabs generally

shawnbot:
Our kebab usage guidelines send mixed messages. In the table at the top of our [progressive disclosure guide](https://primer.style/design/ui-patterns/progressive-disclosure), we say:

> Refrain from using this icon as a progressive disclosure solution

But in the "Ellipses [sic] Icon" section, the first bullet under "Caution" reads (emphasis mine):

> While visually similar, the ellipses icon is different from **the kebab icon, which is used for dropdown menus or general call-to-actions**. Be careful not to confuse them for one another.

This PR simply updates the first part in the table to read:

> For toggling inline dropdowns and menus

Which I think resolves the inconsistency, but I think we could do better by having a separate section about kebabs and dropdown/menus with do and don't screenshots.


emilybrick:
Thanks for calling this out @shawnbot. This PR lgtm. 

When I look at this table, I'm reminded that we're basically telling users all the icons they _can't_ use for dropdowns (in some cases, incorrectly so), without giving them much guidance on where they _should_:
Screenshot 2019-04-18 12 26 04

Also these two (sort of) conflicting statements
Screenshot 2019-04-18 12 26 41
Screenshot 2019-04-18 12 26 46

PR #45, opened 2019-05-20
Now v2 Updates

emplums:
Updates our now.json and creates a nowignore file to be compatible with now v2 🌮 

Still need to figure out the static file situation


colebemis:
Closing this because we'll be refactoring this site with [Doctocat](https://github.com/primer/doctocat) soon. The refactor will include deploying the site with Now v2.

PR #49, opened 2019-06-25
Improve documentation

ZainKho:
Wrote up some documentation for the /design landing page, /design/tools section, and added a couple elements to the Progressive Disclosure section.

### DIFF
--- a/pages/design/index.mdx
+++ b/pages/design/index.mdx
@@ -1,14 +1,16 @@
 # Introduction
 
-The goal of Primer Design is to create an accessible and approachable system that keeps user experience consistent across the breadth of GitHub products and services while also allowing flexibility for new ideas. These guidelines are created with *all* users in mind. These guidelines set a foundation for design principles, accessibility, layouts, content, and resources for multiple design tools.
+Our goal is to create an accessible and approachable system that keeps user experience consistent across the breadth of GitHub products and services while also allowing flexibility for new ideas. These guidelines are created with *all* users in mind. Primer Design sets foundations for design principles, accessibility, layouts, and content. We also provide downloadable resources for multiple design tools.
 
 ## Designed with accessibility first
-We created Primer Design so that GitHub can be used by everyone. We suggest guidelines that create the most opportunity for those with visual, hearing, cognitive, and mobile disabilities.
+We created Primer Design so that GitHub can be used by everyone. Our guidelines have considerations for users with visual, hearing, cognitive, and mobile disabilities.
+
 
 ## Responsiveness
-Primer Design works with all devices and includes guidelines for page layouts, hierarchy, navigation, and so much more.
+Primer Design includes guidelines for page layouts, hierarchy, and navigation on both web and mobile.
+
 
 ## Tools that encourage collaboration
-We use [Figma](http://figma.com) as our primary tool of choice because of its ability to create a more transparent work environment. Figma allows designers to share early and share often.
+We use [Figma](http://figma.com) as our primary tool of choice because of its ability to create a more transparent work environment. Figma supports GitHub's mission to help people work together from all of the world and contribute to amazing projects.
 
 export const meta = {displayName: 'Guidelines', nav: 'top', hero: true}
--- a/pages/design/index.mdx
+++ b/pages/design/index.mdx
@@ -1,3 +1,14 @@
-# Guidelines
+# Introduction
+
+The goal of Primer Design is to create an accessible and approachable system that keeps user experience consistent across the breadth of GitHub products and services while also allowing flexibility for new ideas. These guidelines are created with *all* users in mind. These guidelines set a foundation for design principles, accessibility, layouts, content, and resources for multiple design tools.
+
+## Designed with accessibility first
+We created Primer Design so that GitHub can be used by everyone. We suggest guidelines that create the most opportunity for those with visual, hearing, cognitive, and mobile disabilities.
+
+## Responsiveness
+Primer Design works with all devices and includes guidelines for page layouts, hierarchy, navigation, and so much more.
+
+## Tools that encourage collaboration
+We use [Figma](http://figma.com) as our primary tool of choice because of its ability to create a more transparent work environment. Figma allows designers to share early and share often.
 
 export const meta = {displayName: 'Guidelines', nav: 'top', hero: true}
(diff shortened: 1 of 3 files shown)


pages/design/index.mdx
```diff
@@ -1,3 +1,16 @@
-# Guidelines
+# Introduction
+
+Our goal is to create an accessible and approachable system that keeps user experience consistent across the breadth of GitHub products and services while also allowing flexibility for new ideas. These guidelines are created with *all* users in mind. Primer Design sets foundations for design principles, accessibility, layouts, and content. We also provide downloadable resources for multiple design tools.
+
+## Designed with accessibility first
+We created Primer Design so that GitHub can be used by everyone. Our guidelines have considerations for users with visual, hearing, cognitive, and mobile disabilities.
+
```
ZainKho: Plan to add graphics here.

pages/design/index.mdx
```diff
@@ -1,3 +1,16 @@
...
+
+## Responsiveness
+Primer Design includes guidelines for page layouts, hierarchy, and navigation on both web and mobile.
+
```
ZainKho: Plan to add graphics here.

pages/design/index.mdx
```diff
@@ -1,3 +1,16 @@
...
+
+## Tools that encourage collaboration
+We use [Figma](http://figma.com) as our primary tool of choice because of its ability to create a more transparent work environment. Figma supports GitHub's mission to help people work together from all of the world and contribute to amazing projects.
```
ZainKho: Plan to add graphics here

pages/design/tools/index.mdx
```diff
@@ -1,3 +1,8 @@
 # Design tools
 
+Figma is a vital part of Primer Design's workflow. Figma's collaborative environment allows designers to work together across distance and give feedback asynchronously. Learn more about [GitHub's remote design culture here](https://www.figma.com/blog/github-case-study/).
+
+## Resources
+We provide some open source Figma files to help designers create more with GitHub.com and other GitHub products. Download them below.
+
```
ZainKho: Plan to add Figma file links here

emilybrick: I wonder if it makes more sense for this content to live on https://primer.style/design/tools/figma? Rather than on the main tools page

pages/design/ui-patterns/progressive-disclosure.mdx
```diff
@@ -10,10 +10,10 @@ The following table outlines common progressive disclosure solutions in use at G
...
-Text-only Toggles | | | Usage of this solution is discouraged, as generally icons or icon+text pairings provide better accessibility and more information |
+| Chevron | ![chevron](https://user-images.githubusercontent.com/24916540/52089727-1ae9d480-2564-11e9-8291-1ee9d3225427.png) | When elements of content are collapsed and can be toggled open | Do not use this icon to trigger a dropdown menus or navigation; the Caret icon is the more suitable for this. |
+Expand/ Collapse Icon | ![expand](https://user-images.githubusercontent.com/24916540/52089724-1a513e00-2564-11e9-84f8-eba077f6abfc.png) | Signify that there is content (typically text) to be revealed | Should generally stand alone, rather than being paired with text |
+Ellipses Icon | ![ellipses](https://user-images.githubusercontent.com/24916540/52089725-1ae9d480-2564-11e9-8d29-1dc28740924a.png) | For toggling truncated inline text content | Do not use this icon to trigger a dropdown menus or navigation; the Kebab icon is the more suitable for this. |
+Text-only Toggles | [example]() |  | Usage of this solution is discouraged, as generally icons or icon+text pairings provide better accessibility and more information |
```
emilybrick: Thanks for adding this. There are a few things I'm wondering about here – 

- _Do_ we generally discourage them? Note [this](https://www.dropbox.com/s/gq5di6wy11fbi62/Screenshot%202019-06-26%2011.31.35.png?dl=0) on the dashboard page
- Should we include text-only toggles if we generally discourage them? If we're aligned that this is *not* a pattern we want to encourage, I don't think we should include this at all in the table. 🤔 
- If we end up keeping text-only toggles in this table, it might make sense to use the text 'Show more' instead of 'example' just to make it a bit more clear

ZainKho: Agreed. I meant to ask you where else I could find text toggles. It seems like the only place I can find this is on the dashboard there. If we can find a few more instances, maybe we can get a better idea.

ashygee: I don't know if we need to include this example as it doesn't fully represent the full interaction. Maybe we could show a bit more to get more context?

pages/design/tools/index.mdx
ashygee: +1 for @emilybrick's suggestion

pages/design/tools/index.mdx
```diff
@@ -1,3 +1,8 @@
 # Design tools
 
+Figma is a vital part of Primer Design's workflow. Figma's collaborative environment allows designers to work together across distance and give feedback asynchronously. Learn more about [GitHub's remote design culture here](https://www.figma.com/blog/github-case-study/).
```
ashygee: If we're going to talk about design tools we use I think adding Figma is good but if we are going to talk about design tools we provide we should include prototyping too.

pages/design/tools/index.mdx
ZainKho: I think the navigation for Sketch/Figma is a little out-of-the-way, so I think providing direct download links on the front page encourage a little more action. The Figma/Sketch pages can also host them but maybe focus more on the documentation of each component and principles of using them within the files?

This would require a lot more documentation on the design tool pages, but I think it would help give a better purpose to each page.

pages/design/index.mdx
```diff
@@ -1,3 +1,16 @@
...
```
ashygee: - `https://www.figma.com/`
- I think that both sentences say the same thing. Maybe we can combine into one?

pages/design/tools/index.mdx
ZainKho: Added a sentence about prototyping in the design tools intro text.

pages/design/ui-patterns/progressive-disclosure.mdx
ZainKho: I think it would be great if we could show something like [what Emily suggested](https://www.dropbox.com/s/gq5di6wy11fbi62/Screenshot%202019-06-26%2011.31.35.png?dl=0) in this section.

emilybrick: Just chatted with @emplums about this – let's remove it from the table unless we think there's a valid reason to keep it

PR #60, opened 2019-08-13
fixed security vulnerabilities

ashygee:
fixed lodash security vulnerabilities


jonrohan:
did you need to update package.json also?

ashygee:
fixed in doctocat migration

PR #61, opened 2019-08-14
migrated to doctocat and removed old pages directory

ashygee:
Moving to doctocat! 🎉 
- Renamed from `Primer Design` to `Interface guidelines` for consistency with primer.style
- Removed old docs
- Migrated everything Doctocat


src/@primer/gatsby-theme-doctocat/components/hero.js
```diff
@@ -0,0 +1,22 @@
...
+import {Container} from '@primer/gatsby-theme-doctocat'
+import heroIllustration from '../../../hero-illustration.svg'
+
+const HeroAnimation = loadable(() => import('../../../hero-animation'), {
+  fallback: <img src={heroIllustration} alt="Interface Guidelines hero" width="100%" />
+})
+
+export default function Hero() {
+  return (
+    <Box bg="black" p={5}>
```
colebemis: `Container` already has some horizontal padding so we only need to set `py`.

```suggestion
    <Box bg="black" py={6}>
```

src/@primer/gatsby-theme-doctocat/components/hero.js
```diff
@@ -0,0 +1,22 @@
...
+      <Container>
+        <Heading color="blue.4" fontSize={7} pb={3} m={0}>
```
colebemis: I notice that the line height was a little loose when this heading wrapped on mobile.

```suggestion
        <Heading color="blue.4" fontSize={7} lineHeight="condensed" pb={3} m={0}>
```

PR #99, opened 2020-04-15
Updated design guidelines content and IA proposal

yaili:
- Updated to reflect more recent conversations
- Moved some sections that are likely to be done later to separate section

### DIFF
--- a/docs/content.md
+++ b/docs/content.md
@@ -33,7 +33,7 @@
 * Search and filter
 * Tables and lists
 
-### Worflows
+### Workflows
 
 * Creating
 * Feature onboarding
--- a/docs/content.md
+++ b/docs/content.md
@@ -7,6 +7,7 @@
 * Accessibility: consider a11y in design early, list of tools
 * Design principles
 * Designing for safety
+* Content
 
 ### Foundations
 
--- a/docs/content.md
+++ b/docs/content.md
@@ -69,7 +69,7 @@
 
 * Native
   * Desktop
-  * Atom
+  * Mobile
   * IDEs
 * Workflows
   * Invitations (new org, repo, etc.)
--- a/docs/content.md
+++ b/docs/content.md
@@ -2,12 +2,10 @@
 
 ## Guidelines
 
-###  Global
+###  Principles
 
-* Accessibility: consider a11y in design early, list of tools
-* Design principles
-* Designing for safety
-* Content
+* Accessibility: (Inclusive) consider a11y in design early, list of tools
+* Safety: designing for safety
 
 ### Foundations
 
@@ -15,6 +13,7 @@
 * Color: palette, use/brand (use more, use less), a11y (recommended combos), accent color
 * Spacing: scale, density on product vs spacious on marketing pages
 * Typography: scale, use (bold, normal, light, sans, mono) semantic markup/a11y
+* Content
 
 ### Layout
 
--- a/docs/content.md
+++ b/docs/content.md
@@ -30,7 +30,6 @@
 * Loading states
 * Messaging
 * Progressive disclosure: show more, expand, etc.
-* Search and filter
 * Tables and lists
 
 ### Workflows
@@ -38,6 +37,7 @@
 * Creating
 * Feature onboarding
 * Saving
+* Seaching and filtering
 
 ### Communication
 
--- a/docs/content.md
+++ b/docs/content.md
@@ -1,4 +1,4 @@
-# Design Guidelines content & IA
+# Interface Guidelines: proposed IA
 
 ## Guidelines
 
--- a/docs/content.md
+++ b/docs/content.md
@@ -4,83 +4,78 @@
 
 ###  Global
 
-* Design Principles
-* Accessibility
-* Voice & tone (off link to Manual of Style?)
-* Designing for Safety
+* Accessibility: consider a11y in design early, list of tools
+* Design principles
+* Designing for safety
 
-### Foundation
+### Foundations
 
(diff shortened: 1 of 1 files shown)


docs/content.md
```diff
@@ -4,83 +4,78 @@
...
-* Animation
-* Progressive disclosure (show more, expand etc.)
-* Defensive design (valley of safety vs destructive)
+* Messaging
+* Progressive disclosure: show more, expand, etc.
+* Search and filter
+* Tables and lists
 
-### Layout
+### Worflows
```
broccolini, line 36:
```suggestion
### Workflows
```

docs/content.md
```diff
@@ -4,83 +4,78 @@
 
 ###  Global
 
-* Design Principles
-* Accessibility
-* Voice & tone (off link to Manual of Style?)
-* Designing for Safety
+* Accessibility: consider a11y in design early, list of tools
```
broccolini, line 7: In future I think would be good have "inclusive design" guidelines so as to go broader than accessibility standards.

docs/content.md
```diff
@@ -4,83 +4,79 @@
 
 ###  Global
```
ashygee, line 5: Would "Universal" work here? I see the connection for "Global" (e.g. global styles in CSS) but I think Universal as the category carries more of a "these are general guidelines to use throughout the system" without feeling a connection to development only.

docs/content.md
```diff
@@ -4,83 +4,79 @@
...
-* Accessibility
-* Voice & tone (off link to Manual of Style?)
-* Designing for Safety
+* Accessibility: consider a11y in design early, list of tools
+* Design principles
+* Designing for safety
+* Content
 
-### Foundation
+### Foundations
```
ashygee, line 12: Would we want to include an overview for this section to connect with https://github.com/primer/primitives?

docs/content.md
```diff
@@ -4,83 +4,79 @@
...
+* Defensive design: valley of safety vs destructive, destructive actions, danger
 * Empty states (blankslates)
-* Saving models
 * Loading states
-* Animation
-* Progressive disclosure (show more, expand etc.)
-* Defensive design (valley of safety vs destructive)
+* Messaging
+* Progressive disclosure: show more, expand, etc.
+* Search and filter
```
ashygee, line 33: Would this section include guidelines around filtering vs sorting?

docs/content.md
```diff
@@ -4,83 +4,79 @@
...
+* Saving
 
 ### Communication
 
+* Data formats
+* Emoji
+* Iconography
+* Imagery
+  * Octocats
+  * Illustrations
```
ashygee, line 49: Does this refer to technical illustrations like the images we use for "Feature preview" 

e.g.


or to the octocat illustrations seen across the marketing pages and used in blankslates?

docs/content.md
```diff
@@ -4,83 +4,79 @@
...
+* Codepen
+* Playroom
 
-* Current templates
-* Info re use and maintenance mode
+## Other/future sections
 
+* Native
+  * Desktop
+  * Atom
```
ashygee, line 71: is this still relevant now that Atom is deprecated?

docs/content.md
```diff
@@ -4,83 +4,79 @@
...
-### Sketch
+* Figma: using components, Figma prototyping, sharing and permissions
+* Codepen
+* Playroom
 
-* Current templates
-* Info re use and maintenance mode
+## Other/future sections
 
+* Native
```
ashygee, line 69: Should we include Mobile as a category?

docs/content.md
yaili: I think either would work. I wonder also if this whole section could be called Principles instead, like in Primer CSS, and we move Content to Foundations? 🤔 @ashygee

docs/content.md
yaili: I can make it so it's more broad, and call it Inclusive design. When speaking to the a11y consultants last week they were more keen on that term too. It would also work well if we rename this section to "Principles".

docs/content.md
yaili: That's a good idea.

docs/content.md
yaili: I'm not 100% sure how to separate this section from tables and lists in a way that people can find what they're looking for easily. Searching and filtering always seems to happen within tables and lists, right? Another thought I just had: in the Workflows section I include things that can be defined with action verbs (Creating, Onboarding, Saving), so maybe Searching (and filtering) are part of Workflows? 🤔

docs/content.md
yaili: TBH I left this section as it was before. From what I can see, this Imagery section is less UI assets and more decoration images, to simplify things a lot. Do you think we should restructure this proposal?

docs/content.md
yaili: Yes, updated.

docs/content.md
```diff
@@ -1,86 +1,82 @@
...
 ## Guidelines
 
-###  Global
+###  Principles
 
-* Design Principles
-* Accessibility
-* Voice & tone (off link to Manual of Style?)
-* Designing for Safety
+* Accessibility: (Inclusive) consider a11y in design early, list of tools
```
emilybrick, line 7: Should this just be "Inclusivity": ? Confused about why inclusive is in ()

yaili: I'm not sure the term "Inclusivity" is as widely used as "Inclusive", but I agree that we should try to call it "inclusive design" (once we find the correct/best term).

PR #106, opened 2020-05-28
Transient Components guide

simurai:
Based on this [Slack convo](https://github.slack.com/archives/GACAW0NPM/p1589321694228900) with @vdepizzol + @emplums I got reminded of this draft about our "transient" components.

It would still need some work. Maybe change the "transient" name and it might be already out of date. But could be helpful since the difference between these components can be blurry.


yaili:
I made a few tiny edits. Regarding the name, I don't feel transient is the best title, as that sounds more like a toast or alert, that will disappear without the user making an action. I'm thinking more of words like `reveal`, `disclose` (although we already have [progressive disclosure](https://primer.style/design/ui-patterns/progressive-disclosure), but maybe it's another type of disclosure in this case?). Terms like `interactive` may be too broad, too. 🤔

simurai:
> Regarding the name, I don't feel transient is the best title

Ahh.. yes. I had to look up the meaning again to know what transient means. 🤦 

Or maybe "overlay component"? A component that gets temporarily overlaid.

simurai:
Let's close this again since a lot of these components might get replaced by: https://github.com/github/design-systems/issues/1021

PR #107, opened 2020-06-22
Clarify that non-text elements can span entire width

yaili:
Clarified "Don't" in the [Full width](https://primer.style/design/ui-patterns/messaging#full-width) section

**Before:**
"Don't let the text span the entire width of the window"

**After:**
"Don’t let text span the entire width of the window (but non-text elements can still span the entire width)"

### DIFF
--- a/content/ui-patterns/messaging.mdx
+++ b/content/ui-patterns/messaging.mdx
@@ -118,7 +118,7 @@ Flash alerts that affect every page (e.g. a global message) should use the [full
       <img src="https://user-images.githubusercontent.com/586552/63046881-46e5fb00-bea1-11e9-87ee-80dbeb0bea1c.png" />
     </Box>
     <Text fontSize="0" color="gray.5">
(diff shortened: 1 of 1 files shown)


content/ui-patterns/messaging.mdx
```diff
@@ -118,7 +118,7 @@ Flash alerts that affect every page (e.g. a global message) should use the [full
       <img src="https://user-images.githubusercontent.com/586552/63046881-46e5fb00-bea1-11e9-87ee-80dbeb0bea1c.png" />
     </Box>
     <Text fontSize="0" color="gray.5">
-      Don't let the text span the entire width of the window
+      Don’t let text span the entire width of the window (but non-text elements can still span the entire width)
```
emilybrick, line 121:
```suggestion
      Don’t let a single text block span the entire width of the window (non-text elements can still span the entire width, such as left-aligned text and right-aligned buttons)
```

PR #110, opened 2020-07-13
Save pattern documentation

superbryntendo:
cc @anthonyvsshark @mamuso @yaili


superbryntendo:
first comment (which mentioned people) not displaying correctly.

cc @anthonyvsshark @yaili @mamuso

superbryntendo:
This spec is still in a somewhat early state, but collected some add'l considerations to be addressed:

Note: This guidance is dependent on Toasts accessibility project that @anthonyvsshark and I are already digging into, but following the example of the "Messaging" guidance that has already shipped.

1. Expand rules to include UI + copy do's/don'ts (@anthonyvsshark)
2. Optimistic UI/toast combo has issues when user routes away prior to response due to lack of a persistent toast backend, so we may need to block user interactions pending server response in some cases (@mamuso)
3. Give strongly-opinionated, case-specific guidance for toasts vs flash alerts going forward (@mamuso)
4. Include suggestions for when brand and/or whimsy are appropriate - dependent on conversations with creative (@mamuso)

content/ui-patterns/saving.mdx
```diff
@@ -0,0 +1,161 @@
...
+    - [Example code]
+  - Optimistic update
+    - For interactions where the entity already exists and we merely need to update a field on that object, we should fire off the request to the server, immediately display the desired changes as if they've already happened, and upon response from the server, either cache the result of a successful change to save loading time later or present a clear, concise error message if an error has occurred. If the change will not be obviously visible in the UI the user is currently viewing, we should fire off a [Toast]() to let the user know that the change occurred successfully.
```
yaili, line 16: Is there a better term for "fire off"?

superbryntendo: what do you think of "trigger"? maybe "initiate"? I try to follow the rules of good prose and not repeat the same word *too* often, so if you've got suggestions, I'm open!

content/ui-patterns/saving.mdx
```diff
@@ -0,0 +1,63 @@
...
+  const feature = query(form, '.js-feature-preview-slug', HTMLInputElement).value
+  fire(form, `feature-preview-enroll:${feature}`)
+})
+
+```
+  
+In a few specific cases, such as updating notification preferences, we should use the `data-disable-with` property to indicate that the response is still being sent to signal to the user that they shouldn't navigate away from the current page which may cause the request to fail silently in the event of an error.
+
+```html
+  <input type="submit" data-disable-with="Submitting...">Submit</input>
```
mamuso, line 62: When is it ok to use a different string in `data-disable-with`? If the request is fast, the effect is a bit harsh.

CleanShot 2020-07-16 at 17 32 24

content/ui-patterns/saving.mdx
AnthonyVsShark: I like initiate.

content/ui-patterns/saving.mdx
superbryntendo: That's not something we can control though, so we should support the worst-case scenario, right?

superbryntendo: I believe `data-disable-with` always requires a string, but it probably makes sense to always use the present-tense form of the initial verb?

content/ui-patterns/saving.mdx
```diff
@@ -0,0 +1,63 @@
+---
+title: Saving
+---
+
+GitHub's save patterns help developers to store and update their configurations for different features throughout GitHub's various interfaces. These changes should be represented in the UI accurately, quickly, and obviously. Their behavior should be easy to reason about and inspire confidence and trust.
```
yaili, line 5: developer or users? 🤔

content/ui-patterns/saving.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+All configuration flows should be punctuated with CTAs that feature a logical term such as "Create", "Save", "Delete", or "Update" (e.g. "Create" a new repository, "Add" an SSH key to your profile, "Save" security preferences, "Update" a repository's description, "Delete" an old email address).
```
yaili, line 7: Are these 2nd and 3rd paragraph the TLDR of the rest of the doc? I'm thinking if we should go into detail/guidelines in the intro area, or leave it more "intro-y". I can explain better if needed 😄

content/ui-patterns/saving.mdx
```diff
@@ -0,0 +1,63 @@
...
+title: Saving
+---
+
+GitHub's save patterns help developers to store and update their configurations for different features throughout GitHub's various interfaces. These changes should be represented in the UI accurately, quickly, and obviously. Their behavior should be easy to reason about and inspire confidence and trust.
+
+All configuration flows should be punctuated with CTAs that feature a logical term such as "Create", "Save", "Delete", or "Update" (e.g. "Create" a new repository, "Add" an SSH key to your profile, "Save" security preferences, "Update" a repository's description, "Delete" an old email address). 
+
+Individual settings that are not changed via discrete flows should minimize the required UI by saving configuration changes as they are made (e.g. changing your primary email address, adding an assignee to an issue, unsubscribing from a notification type, following a user, watching a repository).
+
+## There are two primary patterns that should be used for saving user input on GitHub:
```
yaili, line 11: Let's rephrase so the heading isn't a sentence, maybe "Primary patterns" with a sentence explaining below?

content/ui-patterns/saving.mdx
```diff
@@ -0,0 +1,63 @@
...
+### Save and route
+This pattern is primarily used for creation and deletion of new entities such as repositories, organizations, or user accounts. At the end of the flow, the user should be routed to the entity they just created. In some cases such as forking or using a template, we may need to hold the user at a waiting page while the entity is generated. In these cases, we should make the user feel confident that something is happening and that their configuration details will not be lost. In some cases, such as deleting a repository, it may make sense to be explicit that the interaction has occurred by using a [Flash alert](https://primer.style/css/components/alerts).
+
```
yaili, line 14: Will this have a visual example when the visuals are agreed? (same question for optimistic update section)

content/ui-patterns/saving.mdx
superbryntendo: Yup! Good call.

content/ui-patterns/saving.mdx
```diff
@@ -0,0 +1,63 @@
...
+### Optimistic update
+For interactions where the entity already exists and we merely need to update a field on that object, we should initiate an async server request using `remoteForm`, immediately display the desired changes as if they've already happened, and upon response from the server, either cache the result of a successful change to save loading time later or present a clear, concise error message if an error has occurred. If the change may not be obviously visible in the UI the user is currently viewing, we should trigger a [Toast](https://primer.style/css/components/toasts) to let the user know that the change occurred successfully. 
+
+```javascript
```
yaili, line 18: Is this the same across different implementations? Wondering if it's too much implementation detail to go into the design guidelines. Happy to be convinced otherwise.

content/ui-patterns/saving.mdx
superbryntendo: Since the Flash alert page has the actual details, I figured I'd leave it to that? I'm certainly open to it, but I didn't think there was a clear fit. It just felt like any visual detail would be better provided by the components' pages themselves.

If you've got ideas, I'm certainly open to it!

content/ui-patterns/saving.mdx
superbryntendo: It is the same. I think it's important - especially after talking to an engineer I work with - because it seems that many engineers just don't know that these methods are here. My goal is to help the designers and their work inform how engineering builds against this pattern and having this detail feels crucial to that process imo.

content/ui-patterns/saving.mdx
superbryntendo: I thought developers was a nicer term than a generic "users" but that may not be the most correct. I'm open to changing.

content/ui-patterns/saving.mdx
superbryntendo: Updated.

content/ui-patterns/saving.mdx
superbryntendo: Updated.

content/ui-patterns/saving.mdx
mamuso: We definitely can't control the speed of the request, but we can control the copy.

It seems that by recommending a copy change on `data-disable-with` we are opening the door to a lot of fast UI copy jumps.

content/ui-patterns/saving.mdx
mamuso: Do we want to be specific and opinionated about the anatomy of a post-save flash alert? 

Do we always use an icon vs not? What kind of style do we use for success? (depending on the page, you can find default flashes for indicating success)

If we have a point of view, we should document it and provide examples.

content/ui-patterns/saving.mdx
superbryntendo: But then, if it's a long pause and there's no change, that seems worse, right? We could stick to disabling the button, but even that doesn't necessarily declare that something is happening...

AnthonyVsShark: Could we swap in a loading indicator instead of text perhaps?

PR #114, opened 2020-07-31
Inclusive design guidelines

katmeister:
- [ ] Add to missing sections
- [ ] Visuals to supplement

cc @matthewferry @ichelsea @yaili

### DIFF
--- a/content/global/inclusive.mdx
+++ b/content/global/inclusive.mdx
@@ -18,10 +18,10 @@ Going through these questions is most effective within a large brainstorming ses
 
 A great first priming question is, "How can this feature be used to harm someone? How can this feature be misused and create unwanted side effects (**stress cases**)?"
 
-To best understand stress cases as a team, it can be helpful to create specific personas. For example, describe the real world **impact*- of these affect people:
+To best understand stress cases as a team, it can be helpful to create specific personas. For example, describe the real world **impact** of how these affect people:
 - Someone escaping from an abusive relationship
 - Someone experiencing ongoing harassment
-- Someone being dogpiled (multiple people taking the same abusive action against you at once, like commenting)
+- Someone being dogpiled (multiple people taking the same abusive action against another person at once, like commenting)
 
 Take notes from this discussion and update any user journeys accordingly.
 
@@ -31,10 +31,10 @@ Another key question to think about beyond abuse is: "Who might you be excluding
 
 Consider those who are:
 - (Link off to accessibility personas?)
-- Newer to the norms of tech, Software Development, and open source
-- Newer to the Internet
+- Newer to the norms of tech, software development, and open source
+- Newer to the internet
 - Using slower devices
-- On unreliable Internet connections
+- On unreliable internet connections
 - Not fluent in written or spoken English
 
 #### How might we encourage good and positive behavior?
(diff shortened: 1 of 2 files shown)


content/global/inclusive.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+## Starting a new project
+
+As you begin tackling a new design problem with your team, here are a few core considerations to help guide you in your preliminary research.
+
+### Think through key questions
+
+It's important to consider different perspectives from the very beginning, before jumping into any solutions. These initial questions will help you be **intentional and proactive**, rather than reactive later on.
+
+Going through these questions is most effective within a large brainstorming session with your team, where everyone can bring their backgrounds and experiences to the table. You can also continue to ask these questions throughout the design process, especially as scope changes.
```
yaili, line 15: Should we add anything here about teams that work across different timezones, and being mindful of that? For example, a team could have a way for people to add to the discussion asynchronously so that everyone can have a say even if they're not in the room.

content/global/inclusive.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+#### How can this feature be used to harm someone?
```
yaili, line 17: I know the heading before is "Think through key questions", but I wonder if we need to be more explicit that this is where the questions are starting. Would adding: "Question 1" before this heading be too much? Or something similar that makes this clear?

content/global/inclusive.mdx
```diff
@@ -0,0 +1,111 @@
...
+- Someone being dogpiled (multiple people taking the same abusive action against you at once, like commenting)
+
+Take notes from this discussion and update any user journeys accordingly.
+
+#### Who might you be excluding?
+
+Another key question to think about beyond abuse is: "Who might you be excluding from partcipating because of the way the feature is built?" When we excitedly jump quickly into building and shipping, it can be easy to forget how our diverse and international community might view our work.
+
+Consider those who are:
+- (Link off to accessibility personas?)
```
yaili, line 33: Don't forget this link 🔗

content/global/inclusive.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+### Add trust signals
+
+How might we remind people of the humans behind the code?
+- Usage of avatars where appropriate
+- Usage of full names, not just handles, in workplace environments
+- Mutual collaborators
+- Mutual followers/following
+- Shared languages, contribution, repos, orgs
+- Account age (I know this one is tricky)
```
yaili, line 78: I the text in brackets to be removed? Do you mean how long they've had their account?

content/global/inclusive.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+## Preparing to ship
+
+When working on feature landing pages or social media posts, provide diverse stories and feature developers that are representative of the community.
+- [ ] Who are you including? Who are you leaving out?
+- [ ] Would someone from a marginalized background see this content and feel represented?
+
+## After shipping
+
+🚧
```
yaili, line 111: Could we have a final section with a very short list of hand picked resources to read more about this topic, with a tiny blurb for why we picked each? For example the Design for Real Life book from A List Apart came to mind, or Technically Wrong.

Issue #130, opened 2020-12-14
Broken Link on Accessibility Page


starkfire:
Page: https://primer.style/design/global/accessibility

In the [Overview](https://primer.style/design/global/accessibility#overview) section, [508 Reference Guide - 1194.22]() points to a [link](https://www.access-board.gov/guidelines-and-standards/communications-and-it/about-the-section-508-standards/guide-to-the-section-508-standards/web-based-intranet-and-internet-information-and-applications-1194-22) which returns a 404 page.

simurai:
@starkfire Thanks for the report.

I'm not sure what that "508 Reference Guide - 1194.22" used to link to. We might can replace it with https://www.access-board.gov/ict/? Seems the newer version of https://www.access-board.gov/ict/#original508.

PR #131, opened 2020-12-14
Fix "508 Reference Guide - 1194.22" 404 

simurai:
Fixes https://github.com/primer/design/issues/130. I'm not sure if that's exactly the same, but it sounds like a newer version.


ashygee (approved):
Thank you @simurai!

PR #140, opened 2021-01-11
Spelling correction

wbrijesh:


yaili (approved):
Thanks for these corrections, @wbrijesh. I approved all the PRs. Some of the documents are going to be revised soon, but these are still very helpful. 🙏🏼

PR #176, opened 2021-08-18
Fix link

hidde:
The link returned a 404, replacing it with what seems to be the correct path


yaili:
Thank you, @hidde! The link is accessible to GitHub staff only, I edited the PR to clarify instead of changing the link.

yaili (approved):
👍🏼

PR #181, opened 2021-09-13
Broken image links on the ActionList page

maximedegreve:
## Problem
When clicking the preview images in bottom of the [ActionList](https://primer.style/design/components/action-list#options)  page you get redirected to a broken page.

## Solution
Change the relative path to a direct link.
I've also tried to use a relative path but that didn't seem to work on the preview or locally and it's hard to test because links always link to the live version documentation.

## Preview
Screenshot 2021-09-13 at 11 51 48

### DIFF
--- a/content/components/action-list.mdx
+++ b/content/components/action-list.mdx
@@ -199,7 +199,7 @@ Action list items can be selected. Single selections are represented with a [`ch
 
 <Grid gridTemplateColumns={['1fr', null, null, null, '1fr 2fr']} gridGap={4}>
     <div>
-        <Link href="/react/ActionMenu">
+        <Link href="https://primer.style/components/ActionMenu">
             <img alt="Action menu" src="https://user-images.githubusercontent.com/293280/123880964-95b3a700-d8f8-11eb-9775-cbaf165207ed.png" />
         </Link>
     </div>
@@ -214,7 +214,7 @@ Action menus are used for disambiguation, navigation, or to display secondary op
 </Box>
     </div>
     <div>
-        <Link href="/react/DropdownMenu">
+        <Link href="https://primer.style/components/DropdownMenu">
             <img alt="Select panel" src="https://user-images.githubusercontent.com/293280/123881067-cdbaea00-d8f8-11eb-98e4-e57c64489308.png" />
         </Link>
     </div>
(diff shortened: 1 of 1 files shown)


content/components/action-list.mdx
```diff
@@ -199,7 +199,7 @@ Action list items can be selected. Single selections are represented with a [`ch
 
 <Grid gridTemplateColumns={['1fr', null, null, null, '1fr 2fr']} gridGap={4}>
     <div>
-        <Link href="/components/action-menu">
+        <Link href="https://primer.style/components/ActionMenu">
```
colebemis, line 202: I don't think the `primer.style/` is necessary for these links.
```suggestion
        <Link href="/react/ActionMenu">
```

colebemis (commented):
Nice catch, @maximedegreve! Just left a minor comment

yaili (changes requested):
Once you update the links based on Cole's review it's 👍🏼

content/components/action-list.mdx
maximedegreve: Updated that @colebemis. Unfortunately still doesn't seem to work locally or on the preview link. I think it's because the React ones are on another site? 

That's why I was worried that `<Link />` wouldn't work either way as it wouldn't be on the same router? I could be super wrong.

Is there another way to test this? I guess if we can't test it before merging we could merge it anyway since it's broken anyway and smoke test it after it's deployed?

content/components/action-list.mdx
```diff
@@ -224,13 +224,13 @@ Action menus are used for disambiguation, navigation, or to display secondary op
 
 Dropdown menus are used for making a single selection among a small list of options. The list is usually predefined with system values, but in some cases it can include user-provided data when those are known not to grow into too many items.
 
-[Primer React implementation](https://primer.style/components/DropdownMenu)
+[Primer React implementation](/react/DropdownMenu)
```
yaili, line 227: This looks like it's going to /design/react/DropdownMenu in the deployment preview. @maximedegreve can you double check that all image and text links here are doing what they should?

maximedegreve: I've switched them back to absolute links as I'm pretty sure it won't work because the router isn't the same between the interface site and the react site. 😅 This risks making mistakes as we can't test it on the deploy previews. Check it now @yaili  it should work now I think. 👀

yaili (approved):
👍🏼

PR #188, opened 2021-10-13
Token component documentation

mperrotti:
I've added design guidelines for using the Token component. This documentation is not ready to merge yet, but I'd still appreciate some feedback.


yaili (approved):
Looks good! But we should also list this in the [Components](https://primer.style/design/components) overview page.

PR #197, opened 2021-11-30
Add content to index pages

yaili:
- added more text to primer.style/design homepage
- edit layout of **UI patterns** to use same grid as **Components** - images missing but not urgent
- edit layout of **Design tools** to use same grid as **Components** - images missing but not urgent
- add content to **Foundations** index

Note: The content I added is very basic, but the goal right now is just to not have dead end, empty pages.

### DIFF
--- a/content/foundations/index.mdx
+++ b/content/foundations/index.mdx
@@ -1,3 +1,36 @@
 ---
 title: Foundations
 ---
+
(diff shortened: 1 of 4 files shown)


ashygee (approved):
Looks good! If you need assistance with images let me know. 🚀

PR #200, opened 2021-12-10
Update button guidelines so items are easier to find

maximedegreve:
During my first responder week we [received a question in Slack](https://github.slack.com/archives/GACAW0NPM/p1639131957147100) (internal only) regarding order/alignments of multiple buttons. Should primary come before a secondary. I couldn't find the info in our docs until @yaili mentioned that it was in the bottom of the buttons document. 

Because there is clear section for alignments in the sidebar and it was at the end of the page just before support it's easy to miss.

I've moved things slightly around to add an `Alignments` section and add `order` and `full-width` buttons in there. Hopefully this will make it easier to find.

I've also exposed `Icon` and `Counter` in the sidebar since this was hidden under `Variation` and was hard to find as well.

Optional: I've also changed `Button types` to `types` and `Button groups` to `Groups`. I think it's easier to read and button doesn't add any value to this and just makes the sidebar harder to read.

better

### DIFF
--- a/content/ui-patterns/button-usage.mdx
+++ b/content/ui-patterns/button-usage.mdx
@@ -58,6 +58,8 @@ In certain scenarios, buttons require supporting text to give the user more cont
 
 ## Alignment
 
+### Order
+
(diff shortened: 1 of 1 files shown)


yaili (approved):
👍🏼

PR #241, opened 2022-03-22
Add primer-reviewers to CODEOWNERS

siddharthkp:
```diff
+ *                       @primer/reviewers
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
 /content/accessibility/ @primer/accessibility-reviewers
+*                       @primer/reviewers


.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
-/content/accessibility/ @primer/accessibility-reviewers
\ No newline at end of file
+/content/accessibility/ @primer/accessibility-reviewers
+*                       @primer/reviewers
```
jonrohan, line 2: reviewers is everyone on the team, do we need that many?

yaili: what do we do for other repos?

siddharthkp: would be nice to narrow it down, should we create a "design-reviewers" team?

siddharthkp: > what do we do for other repos?

We have `css-reviwers`, `react-reviewers`, etc: https://github.com/orgs/primer/teams

yaili (approved):

PR #232, opened 2022-03-11
chore: Add CODEOWNERS

smockle:
This PR adds @primer/accessibility-reviewers as an owner of https://github.com/primer/design/tree/main/content/accessibility, so that the team can review additions/updates to the accessibility-related docs there.

PR #244, opened 2022-03-22
Fix CODEOWNERS for content/accessibility

siddharthkp:
hat tip @jonrohan

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-/content/accessibility/ @primer/accessibility-reviewers
-*                       @primer/reviewers
+/content/accessibility/* @primer/accessibility-reviewers
+*                        @primer/reviewers


smockle:
Could you please elaborate on why the asterisk was needed in `/content/accessibility/*`? When I initially added this in https://github.com/primer/design/pull/232, I followed the example in https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file:
> # In this example, @doctocat owns any file in the `/docs`
> # directory in the root of your repository and any of its
> # subdirectories.
> /docs/ @doctocat

smockle:
From @siddharthkp in https://github.com/primer/design/pull/244#issuecomment-1075372486:
> You're right, I don't think it's needed. Removed in #246

It looks like #246 also removed the trailing `/`—is that needed?

siddharthkp:
> is that needed?

nope :) works without the `/`: https://github.com/primer/design/pull/247

PR #248, opened 2022-03-22
fix: drop down menu link

jainpawan21:


siddharthkp:
Thanks for taking the time! Don't think we need this anymore after https://github.com/primer/design/pull/236 :)

PR #256, opened 2022-04-07
Fix a typo

cebence:
"without" is making the sentence contradictory.


siddharthkp (approved):
Good catch! Thanks for taking the time

PR #261, opened 2022-04-27
Interface guidelines for forms

mperrotti:
These changes introduce high-level interface guidelines for designing forms.

More specific guidelines for each form-related component will follow in separate PRs.


yaili (changes requested):

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+Labels should be descriptive, and as short as possible: ideally no more than 3 words. Labels should be written in sentence case. For example: "Repository name".
+
+If you're having trouble keeping label text short, consider using a [caption](#caption) to provide more context.
+
+In rare cases, a label may be hidden when the user has enough context to understand what the input does. When a label is visually hidden, label text must still be provided for assistive technology such as screen readers. For example: a search field with an icon that is placed in the spot a user would expect to find a search field may have a visually hidden label that says "Search".
+
+Placeholder text is never an acceptable substitute for a form control label because:
+
+- The placeholder text disappears as soon as the input has a value
+- Placeholder text colors are typically too light to meet the minimum color contrast ratio required to be accessible
```
camertron, line 30:
```suggestion
- Placeholder text colors are typically too light to meet the minimum color contrast ratio required for accessibility
```

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+### Label (required)
+
+Labels should be descriptive, and as short as possible: ideally no more than 3 words. Labels should be written in sentence case. For example: "Repository name".
+
+If you're having trouble keeping label text short, consider using a [caption](#caption) to provide more context.
+
+In rare cases, a label may be hidden when the user has enough context to understand what the input does. When a label is visually hidden, label text must still be provided for assistive technology such as screen readers. For example: a search field with an icon that is placed in the spot a user would expect to find a search field may have a visually hidden label that says "Search".
+
+Placeholder text is never an acceptable substitute for a form control label because:
```
camertron, line 27:
```suggestion
Placeholder text is never an acceptable substitute for a label because:
```

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+  </Do>
+  <Dont>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065188-35462873-3706-4063-b7b9-9d0993665307.png" />
+    <Caption>Don’t use a placeholder as a substitute for a label</Caption>
+  </Dont>
+</DoDontContainer>
+
+### Required field indicator
+
+When a field is required to have a value, it should be visibly marked as required. An individual checkbox or radio cannot be marked as required.
```
camertron, line 46:
```suggestion
When a field is required to have a value, it should be visibly marked as required. An individual checkbox or radio button cannot be marked as required.
```

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+For more information, see the [required fields](#required-fields) section.
+
+### Input (required)
+
+The input is what the user interacts with to set the value of the input. You may pre-fill inputs with smart default values, but be careful about making too many assumptions about what a user wants or needs.
+
+The input must have a minimum contrast ratio of 3:1 to pass WCAG AA-compliance. This is defined in [WCAG success criterion 1.4.11](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html).
```
camertron, line 54:
```suggestion
The input must have a minimum contrast ratio of 3:1 to pass WCAG AA-compliance. Contrast requirements are defined in [WCAG success criterion 1.4.11](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html).
```

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+        alt="valid repo name success"
+        src="https://user-images.githubusercontent.com/2313998/164065221-7e4f6a21-e9a7-465b-9a86-83c61533fdec.png"
+    />
+    <Box as="p" mt="0"><strong>Success messages:</strong> A success message may be used when a user might need extra assurance that the field's value is valid. For example: when creating a repository name, it’s nice to be assured that the name is available and valid.</Box>
+  </Box>
+</Box>
+
+### Validation on submit
+
+The default behavior of the web is to perform validation when the user attempts to submit the form. This lets the user flow quickly through the form without interruption, but the user has to wait until submitting the form to know if they’ve entered invalid data.
```
camertron, line 224:
```suggestion
The default behavior of the web is to perform validation when the user attempts to submit the form. This lets the user flow quickly through the form without interruption.
```

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+  </Box>
+</Box>
+
+### Validation on submit
+
+The default behavior of the web is to perform validation when the user attempts to submit the form. This lets the user flow quickly through the form without interruption, but the user has to wait until submitting the form to know if they’ve entered invalid data. 
+
+When the form fails validation, the first invalid input should be focused and scrolled into the viewport.
+
+After a form has failed validation, you may switch to inline validation to provide quicker feedback than if you made the user re-submit the form to re-check validation statuses.
```
camertron, line 228:
```suggestion
After a form has been submitted and failed validation, you may switch to inline validation to provide quicker feedback.
```

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+### Inline validation
+
+Inline validation provides immediate feedback as the user is entering data. For example: when setting a username, you could show a success message if the name is available or an error message if it’s taken.
```
camertron, line 232:
```suggestion
Inline validation provides immediate feedback while the user is entering data. For example: when setting a username, you could show a success message if the name is available or an error message if it’s taken.
```

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+    <img
+        width="456"
+        alt="screen recording of valid repo name input showing loading indicator after blur"
+        src="https://user-images.githubusercontent.com/2313998/164065857-573ca6db-f2c7-49ac-a3cf-bf479a40973d.gif"
+    />
+    <Box as="p" mt="0">If the form control’s validation is likely to take more than 1 second, show a loading indicator.</Box>
+</Box>
+
+<Note variant="warning">
+  Inline validation offers immediate feedback, but carefully weigh that with the following drawbacks.
```
camertron, line 262:
```suggestion
  Inline validation offers immediate feedback, but be sure to consider the following drawbacks.
```

camertron (approved):
What a phenomenal amount of work, thank you for writing all this up! 🎉

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+<DoDontContainer>
+  <Do>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065202-5c3aba0b-5cef-4ef4-9339-7192a3ebf23a.png" />
+    <Caption>Default to vertically stacked form controls</Caption>
+  </Do>
+  <Dont>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065205-4000430d-2216-4954-a367-86fbd0254cc3.png" />
+    <Caption>Don’t lay out forms into columns just to reduce the vertical space used by the form</Caption>
+  </Dont>
+</DoDontContainer>
```
ashygee, line 134: @mperrotti is the text formatting for the radio selects and the label above the radio select group a new pattern? It feels like the bolding is switched and the group labels (e.g. `Preferred Primer component interface`) should be bold and the radio select options are normal weight.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+### Validation message
+
+A single validation message may be displayed to provide helpful information for a user to complete their task. For example: explain an invalid value and submit the form.
+
+An individual checkbox or radio should not have its own validation.
+
+<DoDontContainer>
+  <Do>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065196-b0146ade-eccb-4a71-b1b6-f4a82b96c115.png" />
```
ashygee, line 83: Are we supporting danger variants for checkboxes now?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+    Primer's form design guidelines aim to minimize the effort and cognitive load required to complete a task that involves a form.
+</Box>
+
+## Form control anatomy
+
+![diagrams labeling the anatomy of a text field and a checkbox field](https://user-images.githubusercontent.com/2313998/164065184-e3c615a3-fb68-4878-b7f8-9a57efb748cb.png )
+
+### Label (required)
+
+Labels should be descriptive, and as short as possible: ideally no more than 3 words. Labels should be written in sentence case. For example: "Repository name".
```
vdepizzol, line 21: > Labels should be descriptive, and as short as possible

"As short as possible" feels like we prioritize sizing over semantics or meaning.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+If you're having trouble keeping label text short, consider using a [caption](#caption) to provide more context.
+
+In rare cases, a label may be hidden when the user has enough context to understand what the input does. When a label is visually hidden, label text must still be provided for assistive technology such as screen readers. For example: a search field with an icon that is placed in the spot a user would expect to find a search field may have a visually hidden label that says "Search".
```
vdepizzol, line 25: > In rare cases, a label may be hidden

> For example: a search field with an icon that is placed in the spot a user would expect to find a search field may have a visually hidden label that says "Search".

Is a search field part of a forms pattern? Search fields appear throughout dotcom, and they're not rare =).

content/ui-patterns/forms.mdx
vdepizzol: Why is this guideline here? Shouldn't the input component provide the correct colors by default?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+A caption may be used to provide additional context about the field to help users fill in the correct data or explain how the data will be used. Caption text should be as short as possible.
+
+Caption text may be displayed with a validation message, or it may be hidden if it is redundant information.
+
+Caption text may be used to augment the [label](#label), but should not be redundant with the label or any other parts of the form control. If the caption feels redundant, try removing it. 
+
+<DoDontContainer>
+  <Do>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065190-5ef9d8db-c9ca-4aec-a2ea-321a81cc080e.png" />
```
vdepizzol, line 66: The caption example feels off here:

> "Set automatically" will determine the closest available region based on your location (IP address)

Can we include something that fits closer with our [Content guidelines](https://primer.style/design/foundations/content)?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+    <Caption>Use caption text that shows new information and provides helpful context</Caption>
+  </Do>
+  <Dont>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065192-f1fd1c28-2031-4bb1-836b-607d14a3d4f4.png" />
+    <Caption>Don’t use caption text that is redundant</Caption>
+  </Dont>
+</DoDontContainer>
+
+### Validation message
```
vdepizzol, line 75: The examples using checkbox groups have some weird application of bold. Do we really want to use bold labels for every checkbox or radio button item? Are the validation messages also meant to be bold? 🤔

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+  </Do>
+  <Dont>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065194-8f123836-5cda-4eea-8f0f-f1c016aa197c.png" />
+    <Caption>Don’t show a repetitive validation message</Caption>
+  </Dont>
+</DoDontContainer>
+
+For more information about form validation, see the [validation guidelines](#invalid-input).
+
+## Input methods
```
vdepizzol, line 107: Shouldn't this section be structured as a list of components for easier parsing, instead of prose with static examples?

[Example](https://primer-6d38263007-26441320.drafts.github.io/components/action-list#application):

> Screenshot of Application section in Action list interface guidelines

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+Show a set of selectable options when there is a finite number of possible values.
+
+## Structure
+
+Forms should have a structure that makes it easy for users to scan. Form controls that are stacked vertically and left-aligned are easier for sighted users to scan visually.
+
+<DoDontContainer>
+  <Do>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065202-5c3aba0b-5cef-4ef4-9339-7192a3ebf23a.png" />
```
vdepizzol, line 127: > 


The radio button and checkbox groups seem to have inconsistent headings and unnecessary bold text. Could we revisit these design decisions?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+- Order fields by their relative level of importance
+- Keep related fields near each other
+- When possible, keep inputs that require keyboard input near each other so that users who use a mouse don't have to switch back and forth between clicking and typing
+
+### Grouping form controls
+
+<Box mb={3} display="flex" alignItems="flex-start" flexDirection={["column", "column", "column", "column", "row"]} sx={{gap: 3}}>
+    <img
+        width="456"
+        alt="visually grouped of related inputs"
```
vdepizzol, line 150: > image

I'm not sure I understand what this image is telling here —
* some headings are inside groups, some are not.
* Is "grouping" specifically referring to horizontal placement?
* Does Group 1.1 inside Group 1 specific which width it should follow?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+        src="https://user-images.githubusercontent.com/2313998/164065207-16e51727-55b2-48bc-930c-ad9f0890ae10.png"
+    />
+    <Box as="p" mt="0">When there is a collection of closely related fields, they should be labeled and visually grouped together. Effective techniques to visually group a collection of form controls include proximity and whitespace.</Box>
```
vdepizzol, line 153: > Effective techniques to visually group a collection of form controls include proximity and whitespace.

Could we be more concise here? :))

In the image, there's a spacing of 8px between elements. I'd make sure we use 16px-equivalent spacing between controls in these scenarios.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+</Box>
+
+### Progressively disclosed form controls
+
+<Box mb={3} display="flex" alignItems="flex-start" flexDirection={["column", "column", "column", "column", "row"]} sx={{gap: 3}}>
+    <img
```
vdepizzol, line 159: > 

Is this a realistic width we should recommend for controls? Everything looks so squished horizontally. Also, the spacing between expiration and CVV feels incorrect. 

Additionally, shouldn't these controls be grouped, as described in the guidelines right above this section?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+<Box mb={3} display="flex" alignItems="flex-start" flexDirection={["column", "column", "column", "column", "row"]} sx={{gap: 3}}>
+    <div>
+        <img
+            alt="Checkbox that progressively discloses a dropdown with an internal label"
+            src="https://user-images.githubusercontent.com/2313998/164065216-a2702413-5e84-4ec0-b990-b72261df0fb1.png"
+        />
+        <img
+            alt="Checkbox that progressively discloses a labeled text input"
+            src="https://user-images.githubusercontent.com/2313998/164065217-7969283a-6dbb-43b6-8f62-e4ab995de246.png"
+        />
```
vdepizzol, line 189: > 

This UI feels confusing, with the checkbox label and text input label having the same alignment and style. I don't think we should use a search icon in that case, but I'm afraid that's not the only thing that's adding to the confusion.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+    <Box as="p" mt="0">If the parent form control does not provide sufficient context, a visible label should be shown. Make sure the label of the nested form control doesn’t clash with the label of its parent form control.</Box>
+</Box>
+
+For more information about progressive disclosure, see [Progressive disclosure](/ui-patterns/progressive-disclosure).
+
+## Validation
+
+### Use Primer for validation
+
+Use Primer instead of browser-native validation UI. Browser-native validation messages are not accessible to screen readers, and they visually clash with Primer styles.
```
vdepizzol, line 200: Do we need this sub-heading?

It feels confusing to read "Use Primer instead of". Instead, we could refer to something like:

"Primer components have built-in validation support (...)"

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+    <Box as="p" mt="0">We shouldn’t attempt to validate an input before the user is done with it. So validation may be performed as the user is typing or making their selection, but only <strong>after the first time the input has been validated</strong> and the <strong>input is in an invalid state</strong>. This gives the user early positive feedback by removing the error if the user makes a change that fixes it.</Box>
+</Box>
+
+<Box mb={3} display="flex" alignItems="flex-start" flexDirection={["column", "column", "column", "column", "row"]} sx={{gap: 3}}>
+    <img
+        width="456"
+        alt="screen recording of valid repo name input showing loading indicator after blur"
+        src="https://user-images.githubusercontent.com/2313998/164065857-573ca6db-f2c7-49ac-a3cf-bf479a40973d.gif"
+    />
+    <Box as="p" mt="0">If the form control’s validation is likely to take more than 1 second, show a loading indicator.</Box>
```
vdepizzol, line 258: > If the form control’s validation is likely to take more than 1 second

Is this the right metric to decide whether to display a loading indicator? What if the process is known as a quick one but the user is on a slow connection?

vdepizzol: Also, if we're extending the height of the input control based on an async request that may take long to process, how do we avoid the content to reflow unexpectedly?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+</Box>
+
+<Note variant="warning">
```
vdepizzol, line 261: I wonder if this note could be simplified a bit, and placed closer to the top of that section before the examples?

vdepizzol (changes requested):
As a general note, we should revisit the alternate text from the images — Instead of saying like "diagrams labeling the anatomy of a text field and a checkbox field", the alternate text should describe what's present in the diagrams or screenshots.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
+---
+title: Forms
```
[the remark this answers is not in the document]
mperrotti: I always forget these. I **will** remember the next one!!

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065192-f1fd1c28-2031-4bb1-836b-607d14a3d4f4.png" />
+    <Caption>Don’t use caption text that is redundant</Caption>
+  </Dont>
+</DoDontContainer>
+
+### Validation message
+
+A single validation message may be displayed to provide helpful information for a user to complete their task. For example: explain an invalid value and submit the form.
+
+An individual checkbox or radio should not have its own validation.
```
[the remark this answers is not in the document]
mperrotti: We don't visually mark an individual checkbox or radio as valid either. How about something like this?

```suggestion
An individual checkbox or radio should not have its own validation message or style.
```

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,295 @@
...
+Information from the caption should not be repeated in the error message. Show the validation message and remove the caption. A validation message makes the field easier to spot when a user is scanning for invalid fields.
+
+<DoDontContainer>
+  <Do>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065196-b0146ade-eccb-4a71-b1b6-f4a82b96c115.png" />
+    <Caption>Hide a caption that is redundant with the validation message</Caption>
+  </Do>
+  <Dont>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065194-8f123836-5cda-4eea-8f0f-f1c016aa197c.png" />
+    <Caption>Don’t show a repetitive validation message</Caption>
```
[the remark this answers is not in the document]
mperrotti: The caption should never change. It should either be shown after the validation message, or hidden.

In this case, the caption sounds repetitive after the validation message. I think this example is a little confusing, so I'll come up with something clearer.

content/ui-patterns/forms.mdx
mperrotti: @vdepizzol - I specifically called this out for when/if we decide to redesign our inputs.

This paragraph may need to be rewritten anyway. I looked at dotcom with all of the inputs having a 3:1 contrast, and it looked way too heavy. We may end up leaning on high-contrast color modes for those inputs.

dotcom test branch: [`mp/control-min-contrast`](https://github.com/github/github/pull/new/mp/control-min-contrast)

content/ui-patterns/forms.mdx
mperrotti: > Is a search field part of a forms pattern?

It _could_ be, but usually it isn't. I was trying to think of the most common and obvious example.

Do you (or any other reviewers) have any better examples?

content/ui-patterns/forms.mdx
mperrotti: These are all meant to be bold. That's how the components are currently designed.

I'd be happy to change the checkbox and radio labels from bold to regular weight in separate PCSS and PRC PRs. What do you think @vdepizzol?

content/ui-patterns/forms.mdx
mperrotti: I really like this. Thanks for the example!

content/ui-patterns/forms.mdx
mperrotti: Yes, let's revisit. I think the checkbox and radio labels should be the same weight as body text.

@vdepizzol - could you open an issue or discussion?

content/ui-patterns/forms.mdx
mperrotti: > some headings are inside groups, some are not.

I think you're referring to the labels in the first three form controls as "headings". None of these are headings, they are input labels.

---

> Is "grouping" specifically referring to horizontal placement?

Horizontal placement is just one of the ways you can visually group two or more inputs.

---

> Does Group 1.1 inside Group 1 specific which width it should follow?

I'm not sure I understand this question.

content/ui-patterns/forms.mdx
mperrotti: +1 for being more concise. Writing is hard 🙈

> In the image, there's a spacing of 8px between elements. I'd make sure we use 16px-equivalent spacing between controls in these scenarios.

So if 16px (`space.3`) is the spacing between closely related inputs, would we use 24px (`space.4`) for the default space between inputs? That seems a little big to me 🤔

content/ui-patterns/forms.mdx
mperrotti: > Everything looks so squished horizontally.

I think everything looks so squished because of the text in the diagram. 

---

> Also, the spacing between expiration and CVV feels incorrect.

It definitely is. My mistake.

---

> Additionally, shouldn't these controls be grouped, as described in the guidelines right above this section?

Yup, another mistake on my part.

content/ui-patterns/forms.mdx
mperrotti: You're right - this isn't a great example. I'll come up with a better way to show a labeled input that is "nested" and progressively disclosed.

content/ui-patterns/forms.mdx
mperrotti: Nah, we don't need it. I think I was just trying to emphasize that we **should not** rely on browser validation.

I'll find a better way to do that :)

content/ui-patterns/forms.mdx
mperrotti: > Is this the right metric to decide whether to display a loading indicator? What if the process is known as a quick one but the user is on a slow connection?

It shouldn't matter what connection the user is on. This is a detail that would be handled in the implementation.

content/ui-patterns/forms.mdx
mperrotti: I'm removing this until we make a decision about that how that WCAG guideline works with our color modes.

content/ui-patterns/forms.mdx
mperrotti: What's rendered here is what's rendered by our components. There has been some discussion below about changing the checkbox and radio labels to the normal weight.

content/ui-patterns/forms.mdx
mperrotti: No

content/ui-patterns/forms.mdx
mperrotti: I updated the image, but it's still a little "squished" because of the annotations.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,393 @@
...
+### Validation message
+
+A single validation message may be displayed to provide helpful information for a user to complete their task. For example: explain why a value is invalid so they can correct it and submit the form.
+
+An individual checkbox or radio should not have its own validation.
+
+<DoDontContainer>
+  <Do>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/164065196-b0146ade-eccb-4a71-b1b6-f4a82b96c115.png" />
+    <Caption>Show a validation message for the group of inputs</Caption>
```
langermank, line 82: Have we considered a validation pattern where if there are multiple errors, we render a single box above the form with a list of each error- which links to each input with the issue? Its my understanding that this is the preferred method for handling multiple form errors, but perhaps that case hasn't come up for accessibility testing? 

This is from the book [Form Design Patterns](https://www.smashingmagazine.com/printed-books/form-design-patterns/)

image from Form Design Patterns book showing an error box above a group of fields, linking to each field that is invalid

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,393 @@
...
+  </Do>
+  <Dont>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/166832269-4756f340-c24a-41bd-8fa9-db807607c874.png" />
+    <Caption>Don’t show a repetitive validation message</Caption>
+  </Dont>
+</DoDontContainer>
+
+For more information about form validation, see the [validation guidelines](#invalid-input).
+
+## Input methods
```
langermank, line 105: Before this is published, it would probably be good to update the focus styles to the new design (this should already exist in Figma today)

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,393 @@
...
+
+### Validating a group of form controls
+
+<Box mb={3} display="flex" alignItems="flex-start" flexDirection={["column", "column", "column", "column", "row"]} sx={{gap: 3}}>
+    <img
+        width="456"
+        alt="checkbox group with an error"
+        src="https://user-images.githubusercontent.com/2313998/164065224-d1ba56ce-c632-4318-b9b2-b75940f771a6.png"
+    />
+    <Box as="p" mt="0">A validation message may be used to describe an error for a group of form controls. For example: if a user checks 3 checkboxes when they’re only allowed to check 2 options.</Box>
```
langermank, line 272: I think my comment above about grouping error messages could be described in this section instead. The case you mention here is a single error message for a group of controls, but we should document how to present multiple individual error messages for a group of controls as well.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,393 @@
...
+</Box>
+
+## Submission
+
+Forms should follow consistent patterns for [submitting and saving data](/ui-patterns/saving).
+
+## Related components
+
+<Box mb={4} display="flex" alignItems="flex-start" flexDirection={["column", "column", "column", "column", "row"]} sx={{gap: 3}}>
+  <img
```
langermank, line 282: I'm not sure an image of ActionMenu without any selection makes sense here- I think if we want to show ActionMenu with selection, the image could include one of those states (multi or single select)

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,393 @@
...
+
+<Box sx={{fontSize: 3}} class="lead" as="p">
+    Primer's form design guidelines aim to minimize the effort and cognitive load required to complete a task that involves a form.
+</Box>
+
+## Form control anatomy
+
+![diagrams labeling the anatomy of a text field and a checkbox field](https://user-images.githubusercontent.com/2313998/164065184-e3c615a3-fb68-4878-b7f8-9a57efb748cb.png )
+
+### Label (required)
```
langermank, line 19: In PRC, I see a prop called `layout` which I think is intended to allow the label to display on top of the input, or on the side (left) before the input. Am I correct in that assumption? The code looks a little broken right now:

image

If this is meant to be inline on the left, is that a pattern we need to support? If yes, we might want to document when you would use an inline label.

content/ui-patterns/forms.mdx
mperrotti: I had considered this pattern, but I tried it on a settings page and it felt like too much error text was being shown at once.

However, this pattern is useful and I think it's worth further exploration. Do you think we could explore it in a separate issue? Or should I do it as part of this PR?

content/ui-patterns/forms.mdx
mperrotti: Great point. We really need to come up with a more future-proof way of rendering pattern examples.

cc @tallys - this is a relevant to the documentation image strategies I've been mentioning lately.

content/ui-patterns/forms.mdx
mperrotti: This prop is meant to be used when you're using a custom input component, and you want the input to be laid out like a checkbox or radio form control. You can see an example in the docs: https://primer.style/react/FormControl#with-a-custom-input

I was stumped by this prop in the component API, as were the other PRC folks I consulted. Do you have any ideas @langermank ? At the very least, we shouldn't show this prop in Storybook stories.

content/ui-patterns/forms.mdx
mperrotti: The semibold error messages are kind of new: https://github.com/primer/react/pull/2013

@colebemis and I think it helps the text feel more balanced with the leading icon, but I'm open to changing it back to regular weight it if it's causing an issue.

Other reviewers - please chime in if you agree with @langermank that validation text should go back to regular weight.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+### Required field indicator
+When a field is required to have a value, it should be visibly marked as required. For more information, see the [required fields](#required-fields) section.
+
+### Input (required)
+The input is what the user interacts with to set the value of the field.
+
+For more specific information about each input, visit the respective component documentation page. For example: [CheckboxInputField](), [TextInputField]()
+
+### Placeholder text
```
alliethu: @mperrotti I mentioned this in the TextInput component review, [but I proposed we completely do away with Placeholder text](https://github.com/github/primer/issues/538#:~:text=a%20loading%20state-,Placeholder%20text,-%F0%9F%94%97). [@jscholes had a similar suggestion](https://github.com/github/accessibility/issues/892#issuecomment-1077764961) as well. 
- https://github.com/github/primer/issues/538#:~:text=a%20loading%20state-,Placeholder%20text,-%F0%9F%94%97

alliethu (commented):
Left a couple comments specifically around placeholders

yaili (approved):

content/ui-patterns/forms.mdx
mperrotti: Thanks for the info. I'm going to remove the "Do / Don't" example so we're not explicitly encouraging the use of `placeholder`.

content/ui-patterns/forms.mdx
langermank: Here's a screenshot from a previous form I worked on:

Contact form with multiple fields showing an error state. At the top of the form, there's a group with a title and list of links to each field, stating what the error is with the field

I think the idea here is if there are multiple errors in a single `<form>`, show them in a labeled group with links to each field. My original comment might have sounded like multiple errors per _field_, but that's not what I mean 😅 that feels problematic in a different way!

@alliethu any thoughts here? I pulled this from a book on accessible forms, and that was a year ago. I haven't researched this recently.

content/ui-patterns/forms.mdx
langermank: Got it, I totally thought this was implying a label inline vs. block.. so its more like a private API. I might have called the prop `labelPosition`? But I haven't thought it through too much 😄 in the meantime, you could always add a description to that prop in Storybook to explain what the does or maybe when to use it.

content/ui-patterns/forms.mdx
alliethu: @langermank Your annotation is very much inline with what I would have done myself. I'm attaching a rough annotation below that includes a few more details: 

Error summary notation

Considerations to make:
- add `role="alert"` to the summary `div`
- move keyboard focus to the error summary
- use a semantic heading (I'm thinking an `h2` to title the error summary area)
- reference the heading in the summary `div` by using `aria-labelledby`
- use a semantic list for the error summaries
- when a link is activated, place focus in its corresponding input field
- the error messages should be worded the same in the error summary and inline with each input.
cc/ @mperrotti

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,382 @@
...
+    Forms are used to complete tasks that require data input from the user. For example: creating a new repo, configuring settings, and logging in.
+</Box>
+
+<Box sx={{fontSize: 3}} class="lead" as="p">
+    Primer's form design guidelines aim to minimize the effort and cognitive load required to complete a task that involves a form.
+</Box>
+
+## Form control anatomy
+
+![diagrams labeling the anatomy of a text field and a checkbox field](https://user-images.githubusercontent.com/2313998/164065184-e3c615a3-fb68-4878-b7f8-9a57efb748cb.png )
```
heyamie, line 17: I was a bit confused at first looking at this example, because my immediate reaction wasn't thinking that this was a vertical vs. horizontal layout, but rather two different form types (text input vs. radio check). Do we need to classify this as horizontal vs. vertical, or is it enough to describe the different parts of the forms (label, input, ...)? 🤔

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,382 @@
...
+
+### Open-ended text
+
+![Text input, open-ended autocomplete text input, plain text input, textarea](https://user-images.githubusercontent.com/2313998/168104196-e0400967-a455-4513-97b0-3005d7ee3d0c.png)
+
+Use an open-ended text field when the field does not have a list of possible values. If the input is able to suggest values, use [autocomplete](#autocomplete) to allow users to pick a value or enter their own.
+
+### A set of selectable options
+
+![Action menu, autocomplete input with limited options, checkbox group, radio group, select](https://user-images.githubusercontent.com/2313998/168104199-2990e43c-6204-4ef5-970c-c9babaf1196c.png)
```
heyamie, line 104: As a consumer of this content, `Action menu` and `Select` look pretty close to me, and I have no context to help me select which version I should pick. I wonder if we could provide additional content, or a link to help us make this decision?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,382 @@
...
+
+Show a set of selectable options when there is a finite number of possible values.
+
+## Structure
+
+Forms should have a structure that makes it easy for users to scan. Form controls that are stacked vertically and left-aligned are easier for sighted users to scan visually.
```
heyamie, line 110: In production, we have a few examples of forms breaking this rule (in `Theme preferences` the radio btns stack horizontally, for example).

The do/don't format reads to me as a _must_ do, whereas it seems like this might be a suggestion instead? 🤔

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,382 @@
...
+        alt="Primary discipline radio buttons, selected 'Other' option reveals a text input"
+        src="https://user-images.githubusercontent.com/2313998/164065214-0c0bc39f-52e5-4534-9c69-a1f6e4566fa9.png"
+    />
+    <Box as="p" mt="0">If the parent form control provides sufficient visual context, you may visually hide the label. However, you must specify text for a visually hidden label that is accessible to screen readers.</Box>
+</Box>
+
+<Box mb={3} display="flex" alignItems="flex-start" flexDirection={["column", "column", "column", "column", "row"]} sx={{gap: 3}}>
+    <div>
+        <img
+            alt="Checkbox that progressively discloses a dropdown with an internal label"
```
heyamie, line 170: This is the first time I'm being introduced to the `internal label` concept, I wonder if this is something we should highlight above, on the `Labels` section?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,382 @@
...
+<Box mb={4} display="flex" alignItems="flex-start" flexDirection={["column", "column", "column", "column", "row"]} sx={{gap: 3}}>
+  <img
+    role="presentation"
+    width="456"
+    src="https://user-images.githubusercontent.com/2313998/166846802-7d3a0563-d853-4d7b-b26f-0bb4e8c0d3ad.png"
+  />
+  <div>
+    <Heading fontSize={3}>Form control</Heading>
+    <p>
+      A form control renders a labelled input and, optionally, associated validation text and/or hint text.
```
heyamie, line 322: This is probably out of scope for this particular task, but I just wanted to raise that if I wanted to have a group of radio inputs, I could pick either `Form control` or `Radio group` 😬 It's not totally clear why I should pick one vs. the other

content/ui-patterns/forms.mdx
langermank: @heyamie I think the goal with interface guidelines is to try and document the ideal pattern even if its not always true today. The theme preference example is a great callout! That code and design will need to be addressed based on this guideline. I wonder if we need some kind of explainer that these do/dont's might not be true in production at the time of documentation 🤔

content/ui-patterns/forms.mdx
mperrotti: Thank you both for the helpful examples! I misunderstood what was being suggested. I thought @langermank was suggesting this pattern for a single group of checkbox or radio inputs.

It feels aggressive to _always_ show the error summary alert banner. We should explain when a summary is needed and when it isn't. Maybe when there are 3 or more errors?

mperrotti: I'm noticing repetitive error messages in @alliethu's example, and that feels like it could be overwhelming or confusing to sighted and non-sighted users.

content/ui-patterns/forms.mdx
mperrotti: Thanks for calling that out. I don't think we need to classify horizontal vs vertical. I see how it's confusing.

content/ui-patterns/forms.mdx
mperrotti: This will be covered in the `Action menu` and `Select` component docs.

content/ui-patterns/forms.mdx
mperrotti: Great point about horizontal radio buttons. The intention of this "do/don't" is to discourage forms that flow into multiple columns for no reason. I think it's totally fine to have horizontal radio buttons or have two related text inputs like "First name" / "Last name" on the same line.

I'll update the "do/don't" images to be more clear.

content/ui-patterns/forms.mdx
mperrotti: I will make sure this is clear when we write the guidelines for those components.

Incase you're curious: the main difference is that a `Form control` is used for a single input, and a `Radio group` is used to group multiple related radio inputs. The markup is rendered differently to support assistive technology for screen readers, and the typographic hierarchy is different. Compare this [RadioGroup example](https://primer.style/react/RadioGroup#with-caption) to this [FormControl example](https://primer.style/react/FormControl#with-a-caption).

content/ui-patterns/forms.mdx
mperrotti: I pushed [a commit](https://github.com/primer/design/pull/261/commits/0c55099ae9f157dbdc2c9857d1e1070638ecbdf6) to add some guidance on designing error summaries. I'm not feeling great about it yet, and I'd appreciate some time pairing with somebody from the accessibility team.

content/ui-patterns/forms.mdx
mperrotti: This is only something we'd do with an `Action menu` component, so that information will go in that component's guidelines.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,387 @@
...
+
+import {Box, Heading, Link} from '@primer/components'
+
+<Box sx={{fontSize: 3}} class="lead" as="p">
+    Forms are used to complete tasks that require data input from the user. For example: creating a new repo, configuring settings, and logging in.
+</Box>
+
+<Box sx={{fontSize: 3}} class="lead" as="p">
+    Primer's form design guidelines aim to minimize the effort and cognitive load required to complete a task that involves a form.
+</Box>
```
vdepizzol, line 13: @mperrotti Can we simplify this leading description and make it into a single paragraph?

> Forms are used to complete tasks that require data input from the user. For example: creating a new repo, configuring settings, and logging in.
>
> Primer's form design guidelines aim to minimize the effort and cognitive load required to complete a task that involves a form.

content/ui-patterns/forms.mdx
vdepizzol: @mperrotti have you thought about this?

How about this?

```suggestion
Labels should be descriptive and concise: aim for no more than 3 words. Write labels in sentence case. Examples: "Repository name", "Payment method".
```

content/ui-patterns/forms.mdx
vdepizzol: @mperrotti I'm not sure. It's confusing to see an example about a search field while this page is specifically about [long-form] forms pattern.

content/ui-patterns/forms.mdx
vdepizzol: @mperrotti I'd keep it at a regular weight. Validation messages can appear multiple times to the user in the same form. The bolded font feels too heavy.

content/ui-patterns/forms.mdx
vdepizzol: @mperrotti have you explored this?

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,387 @@
...
+
+A caption may be used to provide additional context about the field to help users fill in the correct data or explain how the data will be used. Caption text should be as short as possible.
+
+Caption text may be displayed alongside a validation message, or it may be hidden if it only provides redundant information.
+
+Caption text may be used to augment the [label](#label), but should not be redundant with the label or any other parts of the form control. If the caption feels redundant, try removing it. 
+
+<DoDontContainer>
+  <Do>
+    <img role="presentation" src="https://user-images.githubusercontent.com/2313998/166832266-c34634ec-2ade-4bf6-8401-df0bdc02b324.png" />
```
vdepizzol, line 53: image

@mperrotti  This "do" example feels quite redundant 🤔

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,387 @@
...
+    Forms are used to complete tasks that require data input from the user. For example: creating a new repo, configuring settings, and logging in.
+</Box>
+
+<Box sx={{fontSize: 3}} class="lead" as="p">
+    Primer's form design guidelines aim to minimize the effort and cognitive load required to complete a task that involves a form.
+</Box>
+
+## Form control anatomy
+
+![diagrams labeling the anatomy of a text field and a checkbox field](https://user-images.githubusercontent.com/2313998/168928812-fa5c9ad7-4f92-4e9d-8e6e-62eb4413985c.png)
```
vdepizzol, line 17: The anatomy image is very blurry on non-high-resolution displays (same for most images on this page in fact). Compare it with the paragraph text in the screenshot:

> Screenshot

For the ActionList guidelines, I made the screenshots exactly 2x the rendered output. @mperrotti can you make these adjustments?

content/ui-patterns/forms.mdx
vdepizzol: >  image

@mperrotti additionally, is this "Pick your top two" and "Only two selections are allowed" a good recommendation for checkboxes? It feels out of place — I could see that pattern happening for a larger number ("Pick up to 10 items"?), but two...? Feels like a weird pattern...

vdepizzol:
@mperrotti I think we should have a rationale for form widths when in a form composition. It's hard to parse from the examples what the recommendation should be:

| From "Structure" | From "Progressively disclosed form controls" | From "Inline validation" |
|--------|--------|--------|
| image | image | Screen Shot 2022-05-23 at 11 23 54 |

content/ui-patterns/forms.mdx
mperrotti: Good point about these guidelines being for long-form forms. I think it might be best to exclude this paragraph. Visually hidden labels can be addressed in form **component** guidelines.

content/ui-patterns/forms.mdx
mperrotti: Ok, I'll update the example images and submit a PR to PRC to change the weight back to normal.

content/ui-patterns/forms.mdx
mperrotti: > if there's a server request, wouldn't it make more sense to show the loading spinner regardless?

@vdepizzol - I was trying to avoid showing too many loading indicators because they could make the app feel slower. If we wait 1 second, we're still giving the user feedback that something is happening, but only if it's something that's making the user wait a noticeable amount of time.

Does that make sense?

content/ui-patterns/forms.mdx
mperrotti: I thought I removed this example. If it's still around somewhere, I'll change it.

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,387 @@
...
+        alt="a form for biographical information that has three invalid inputs"
+        src="https://user-images.githubusercontent.com/2313998/168926352-c30563ab-3782-4b86-8d21-1ad6dcfa09e3.png"
+    />
+    <div>
+      <Box as="p" mt="0">In a <a href="/ui-patterns/messaging#flash-alerts">flash alert</a> at the top of the form, list the invalid inputs as anchor links. When the link is activated, place focus in it's corresponding input.</Box>
+      <Box as="p" mt="0">When the flash alert appears, it should be focused.</Box>
+    </div>
+</Box>
+
+### Inline validation
```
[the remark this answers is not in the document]
mperrotti: You're right - all this motion is overwhelming. I'm going to try converting these to `<video>` elements that can be played or paused individually.

content/ui-patterns/forms.mdx
mperrotti: Good catch. These aren't screenshots, they're rendered and exported from Figma. I'll try bumping the export up to @2x

content/ui-patterns/forms.mdx
mperrotti: I see this is still used used for a do/don't example.

Screen Shot 2022-05-23 at 5 59 28 PM

content/ui-patterns/forms.mdx
langermank: For this particular image, what do you think about pulling those pointer lines out a bit instead of overlapping elements? It looks a little busy to me. You might also consider aligning the label text on one side– so for the lefthand labels (Label, Input) they could all align on the right (end). 

I have an example handy from the Autocomplete docs:

Autocomplete anatomy

content/ui-patterns/forms.mdx
```diff
@@ -0,0 +1,581 @@
...
+    role="presentation"
+    src="https://user-images.githubusercontent.com/2313998/170069941-0e7ead2f-1e20-47dc-b7b0-e7a27480d2a5.png"
+  />
+  <Box>
+    <Box as="p" mt="0">
+      Form controls should be sized to fit their value. Start with the browser default width, and adjust as needed.
+    </Box>
+    <Box as="ul" pl="1rem">
+      <li>
+        Single-line text inputs get a default width set by the browser, but their width can be changed to fit the
```
langermank, line 174: What is the default width? It might be helpful to mention that here.

mperrotti: We don't define it - It's set by the browser.

content/ui-patterns/forms.mdx
langermank: @mperrotti the connector lines between the label and the component with the circle on the end- the way the circle overlaps part of the component (or whatever its pointing to) makes it a little hard to see throughout. Just wondering if it might look more clear to add some white space between the circle and the component like in the screenshot of Autocomplete?

circle overlapping ui element

mperrotti: Thanks for clearing that up. I think that would be a nice improvement. Will fix.

mperrotti:
All of the requested changes have been made and this has just been sitting here. I'm merging it.

If there are additional changes needed, we can file issues and create new PRs.

PR #262, opened 2022-05-19
Fixing mistake in headline

lukasoppermann:
If I am not mistaking the headline: "Content hover on focus" should actually be "Content on hover or focus" (like the linked resource).

The current title got me slightly confused as I was expecting the section to talk about hover content that is disabled on focus. This hover does not quite fit to the content of the section.


siddharthkp (approved):
Thanks for fixing this!

PR #263, opened 2022-05-19
Adding a missing "to" for better flow while reading

lukasoppermann:


siddharthkp (approved):
Good catch, thank you!

PR #271, opened 2022-06-27
Fix spelling

khiga8:
Fixing a spelling error. My bad!


lindseywild (approved):
I didn't even notice 😅 Good catch!

PR #273, opened 2022-06-30
add erblint-github to recommended linter

khiga8:
updates tooling list to include the Gem we manage which contains several accessibility rules


owenniblock (approved):
👍

PR #276, opened 2022-07-25
Segmented control interface guidelines

mperrotti:


keithamus (approved):
LGTM :+1:

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+      src="https://user-images.githubusercontent.com/2313998/180624831-82387f4d-29ec-4d14-b3d4-edbdc58f6257.png"
+      role="presentation"
+    />
+    <Caption>Don't use obscure icons without a visible label.</Caption>
+  </Dont>
+</DoDontContainer>
+
+### Button label text
+
+Segmented control button labels should be nouns or noun phrases that succinctly describe the choice. Button labels may not wrap to multiple lines.
```
simurai, line 162: 👍 I still have to fix the wrapping in PCSS.

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+    <Text as="h4" fontSize={2} fontWeight="bold" m={0}>
+      Only show icons
+    </Text>
+    <Text as="p" mt="0">
+      If your segmented control uses icons that can be understood without a visible text label, then you can hide the
+      text labels and rely only on icons. The text from the hidden label will still be visible in a tooltip.
+    </Text>
+  </figcaption>
+  <img
+    src="https://user-images.githubusercontent.com/2313998/180624846-a906563a-0878-44ce-89d1-46905b190d20.png"
```
simurai, line 330: This image seems to be repeated from the one [above](https://github.com/primer/design/pull/276/files#diff-28ec7427197ac7b3c80bca64eb9536b47e0dc24cf4e03d2e6a95c9c1b58b1f10R305).

simurai (approved):
Just a few minor comments, but otherwise these guidelines are great. 👏

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+      src="https://user-images.githubusercontent.com/2313998/180624838-1915681f-1c50-4d0e-94e9-98f3537cf285.png"
+      role="presentation"
+    />
+    <Caption>Change or add icons to indicate whether the choice is selected.</Caption>
+  </Dont>
+</DoDontContainer>
+
+<DoDontContainer>
+  <Do>
+    <img
```
vdepizzol, line 124: The half-sized images are looking blurry on non-high-resolution monitors. Preview:

Screen Shot 2022-08-01 at 11 09 20

I've documented in this [Figma link](https://www.figma.com/file/HnxaGzO6HrEe5dgx6m5BsH/%5BArchived%5D-Overlay-patterns-(FY22-Q4)?node-id=3%3A47278) which sizes we should use for documentation images. @mperrotti let me know if you have any questions!! 💖

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column-reverse', 'column-reverse', 'column-reverse', 'column-reverse', 'row']}
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+<div>
+
+A segmented control is treated like a [toolbar](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role), not a [radio group](https://developer.mozilla.org/en-US/docs/web/accessibility/aria/roles/radiogroup_role).
```
vdepizzol, line 370: What's the recommendation when a SegmentedControl is used as a tab (akin to TabNav/UnderlineNav behavior)?

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+</DoDontContainer>
+
+#### Icon-only buttons
+
+Use icons without a visible text label sparingly, and only with icons that are easily understood by an average GitHub user. When a segment has an icon without a visible text label, the text label should be shown in a tooltip.
+
+<DoDontContainer>
+  <Do>
+    <img
+      src="https://user-images.githubusercontent.com/2313998/180624830-b91217ae-7c79-4698-a71a-a482576e5df9.png"
```
vdepizzol, line 146: The hover style shown in this image is looking a little squished. Is that also in the implementation?

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+  mb={3}
+  sx={{gap: 3}}
+>
+  <Box as="figcaption" mt="0">
+    <Text as="h4" fontSize={2} fontWeight="bold" m={0}>
+      Label
+    </Text>
+    <Text as="p" mt="0">
+      Segmented controls must have a text label, even if the label is only accessible to screen readers. However, a
+      visible text label is preferred.
```
vdepizzol, line 198: This is a little confusing. Can we list when a visual label is preferred? There are a number of scenarios where the context/location of the SegmentedControl are enough to understand what's it for.

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+  src="https://user-images.githubusercontent.com/2313998/180624841-cc4e4bf9-098c-425c-ba84-7df8413c2055.png"
+  role="presentation"
+/>
+
+A single-column layout is the same way form controls (for example, text inputs) are layed out. The single-column layout is useful for preserving horizontal space.
+
+#### Two-column
+
+<Box display="flex" mb={3} flexWrap="wrap" sx={{gap: 3}}>
+  <img
```
vdepizzol, line 239: There are a couple of image renderings here showing a shadow that shouldn't be visible outside of the element boundaries.

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+
+A segmented control does not require a "Save" button to apply the selection. When a choice is selected, the selection immediately affects an object, state, or view.
+
+## Layout
+
+### Composing the control with a label and caption
+
+#### Single-column
+
+<img
```
vdepizzol, line 229: This design is showing a two-line toolbar with labels throughout:

image

We don't have this anywhere on the website, and I'm afraid it's a design that is too spacious for toolbars. I think we need to validate how this would work in full-page renderings before recommending this pattern.

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+    intended to make the segmented control easier to adapt into different layouts.
+  </Text>
+  <img
+    src="https://user-images.githubusercontent.com/2313998/180624849-a8d785f2-0a0c-42dd-915c-65e26a09eb64.png"
+    role="presentation"
+    width="456"
+  />
+</Box>
+
+### Adapting to narrow areas
```
vdepizzol, line 279: Change it to "Responsive design"?

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column-reverse', 'column-reverse', 'column-reverse', 'column-reverse', 'row']}
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+  <Text as="p" mt="0">
+    A segmented control's width can stretch each button evenly to fill the width of the control's parent. This option is
+    intended to make the segmented control easier to adapt into different layouts.
```
vdepizzol, line 270: When should we use full-width vs left-aligned?

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+    role="presentation"
+    width="456"
+  />
+</Box>
+
+### Adapting to narrow areas
+
+There are cases where a segmented control needs to be placed in a spot that is too narrow to accommodate all of its options. In these cases, the control can be adapted to fit in the narrow area without sacrificing usability.
+
+<Box
```
vdepizzol, line 283: This `<Box>` is centering text and images vertically, which is producing some weird spacing. Maybe try to align to the top?

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+  display="flex"
+  alignItems="center"
+  flexDirection={['column', 'column', 'column', 'column', 'row']}
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+  <figcaption>
+    <Text as="h4" fontSize={2} fontWeight="bold" m={0}>
+      Convert to an action menu
```
vdepizzol, line 294: How can we make this behavior part of SegmentedControl? It might be the most common responsive adaptation, we should make it simple for whoever is using it :)

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+    </Text>
+    <Text as="p" mt="0">
+      Swapping out the segmented control for an action menu is especially helpful when:
+      <Box as="ul" mt={1} pl="1rem">
+        <li>The icons are cannot be understood without visible text labels</li>
+        <li>The control is stil too wide when text labels are hidden</li>
+      </Box>
+    </Text>
+  </figcaption>
+  <img
```
vdepizzol, line 304: This image has labels such as "Wide viewport" and "Narrow viewport", but "Wide" is 1400px+ of width. Maybe call it "Regular viewport"?

---

Screen Shot 2022-08-01 at 11 31 25

Additionally, it feels like moving the external label to an inline option is very prone to layout problems. See my toolbar comment in the "Single column" section above.

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+  alignItems="center"
+  flexDirection={['column-reverse', 'column-reverse', 'column-reverse', 'column-reverse', 'row']}
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+
+<Text as="p" mt="0">Ensure segmented control buttons have a large enough touch target size. The buttons should respond to hovers, clicks, and taps anywhere in the touch target area, even if it isn't directly on the control.</Text>
+  <img
+    src="https://user-images.githubusercontent.com/2313998/180624850-cd500a01-518a-4f93-81b9-fb38410a4371.png"
```
vdepizzol, line 422: Touch targets need 44x44 min sizes, both for width and height. Icon-only SegmentedControl buttons need to have a `min-inline-size` of `var(--control-minTarget-auto)`.

vdepizzol (commented):
Excited for this! Let a couple of quick comments! 🙇

content/components/segmented-control.mdx
mperrotti: I'm bringing this to a11y office hours today: https://github.com/github/accessibility/issues/1556#issuecomment-1204047337

I'll keep you posted.

content/components/segmented-control.mdx
```diff
@@ -0,0 +1,426 @@
...
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column', 'column', 'column', 'column', 'row']}
+  sx={{gap: 3}}
+>
+  <Text as="p" mt="0">
+    An icon can be rendered in a segmented control button as a visual cue to help users understand the choice that
+    button represents.
+  </Text>
+  <img
```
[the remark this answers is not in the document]
mperrotti: Thanks for flagging that. I'll update it in our Figma tools too.

content/components/segmented-control.mdx
mperrotti: Yes, it's a 32x32 square with a 44px high click area on devices with a 
"coarse" pointer: https://primer.style/react/storybook/?path=/story/segmentedcontrol-examples--icons-only

content/components/segmented-control.mdx
mperrotti: > Can we list when a visual label is preferred?

I think we generally prefer visible labels for everything because it adds clarity. cc @alliethu or @ichelsea to keep me correct or add a better explanation.

---

> There are a number of scenarios where the context/location of the SegmentedControl are enough to understand what's it for.

Great point. I'll make sure to directly mention that in this part of the docs

content/components/segmented-control.mdx
mperrotti: Ok, I'll think of a better example to show a segmented control with the label stacked on top.

Any suggestions?

content/components/segmented-control.mdx
mperrotti: I had a hard time explaining this because I don't have any specific examples yet. Here are two slightly less vague use-cases:
- A designer might choose to use a full-width segmented control to make a cleaner layout where elements align to the left and right edges
- To make sure each button has an even width. If the text of one button is very long, that button would be very wide and have a heavier visual weight than the other buttons.

I think that extra information would just add confusion, and it's better not to try and specify use cases until we have a better example. What do you think?

content/components/segmented-control.mdx
mperrotti: This is already built in to the SegmentedControl React component by using the `variant` prop. This info is mostly for designers who are making mocks in Figma.

content/components/segmented-control.mdx
mperrotti: > Maybe call it "Regular viewport"?

👍 

---

> Additionally, it feels like moving the external label to an inline option is very prone to layout problems.

I included the inline label because it's a valid option for action menus. I agree that it's probably too wide for most cases, but it didn't feel correct to exclude the option entirely.

content/components/segmented-control.mdx
mperrotti: I think you're right, but this didn't come up in the a11y design review. I'll double-check with @ichelsea that we do in fact need the width to be at least 44px.

mperrotti: Ok, here's why this passed a11y design review:

> in WCAG 2.2 AA they’re looking at putting a minimum size of 24px by 24px so it would pass that

content/components/segmented-control.mdx
mperrotti: Oops. Will fix.

content/components/segmented-control.mdx
mperrotti: The recommendation was to not use a segmented control like a tab: https://github.rewatch.com/video/jlbodxdruw5bzc8b-accessibility-office-hours-august-3-2022?t=475.87

cc @simurai

content/components/segmented-control.mdx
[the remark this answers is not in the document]
mperrotti: Our Figma tooling is correct, I just didn't have the latest components in my file 😅

Updating my images...

mperrotti:
I have to update the keyboard navigation video based on feedback from @jscholes. During a11y office hours, I learned that a segmented control's keyboard navigation should be like a list (use tabs to focus buttons), not a toolbar (use arrows to focus buttons).

content/components/segmented-control.mdx
simurai: @mperrotti Thanks for bringing this up in the a11y office hour.

> The recommendation was to not use a segmented control like a tab

As I understood it, `SegmentedControl` should behave like a list of items where each `tab` keypress moves to the next item and you can make it active by pressing `enter` or `space`.

But switching between "Write/Preview" in the comment box should use what they call a "composite widget". Where only one tab-stop exists and then you would use the arrow keys to toggle between "Write" and "Preview". Basically keep the current [TabNav](https://primer.style/view-components/components/alpha/tabnav). Keep using tabs should also make it more clear that the two options are mutually exclusive.

PR #284, opened 2022-08-30
Updating color docs

lukasoppermann:
## Changes:
- added reason for why we recommend designing in light mode + plugin link [L21-24](https://github.com/primer/design/compare/update-color-docs?expand=1#diff-68de77b02787c6f3ce1e2201d79c577c5fa60d2e5bd3a3179f44590149d5d02aR21-R24)
- removed usage of _variables_ in favor of design tokens, using one term only makes it easier to understand. Variables are one way to use tokens, js properties are a different one. In this section we discuss the core concept so sticking to tokens seems correct.
- added short explanation what design tokens are good for
- updated example to use only primitives, as we have not introduced component tokens at this point in the docs.
- added section about component tokens, as they exist as well

### DIFF
--- a/content/foundations/color.mdx
+++ b/content/foundations/color.mdx
@@ -17,17 +17,26 @@ import {Box, Button, Heading, Label, LabelGroup, Link} from '@primer/react'
   src="https://user-images.githubusercontent.com/6951037/146927448-cb518377-114a-4ab8-a37e-1eeffc8732c7.png"
 />
 
-GitHub's UI offers a variety of different color modes. When designing product interfaces, you should design in light mode by default. Every pattern in Primer is built to work across all color modes out of the box.
+
+GitHub's UI offers a variety of different color modes. Every pattern in Primer is built to work across all color modes out of the box.
+
+When designing product interfaces in Figma, we recommend to design in light mode. This is best, because the primer components for Figma are only available in light mode.
+To preview your work in other modes, use the [Figma color mode plugin](https://www.figma.com/community/plugin/992128487074360945/Change-Color-Mode).
+
+## How to use color for product UI
+
+Primer delivers colors in the form of [design tokens](https://primer.style/primitives/colors). Design tokens are a layer of abstraction that allows better maintainability, consistancy and easy theming.
+
+For example you use `bg-default` for the background of the page and `fg-default` for the text color. If the user changes to a dark mode, we change the underlying color that those tokens reference, but the tokens stay the same. 
 
 <img
   width="960"
   alt="Same component shown in light mode and dark mode color palettes"
-  src="https://user-images.githubusercontent.com/6951037/127494554-07b67dbd-66d3-4d36-8350-5ec837d6d1a5.png"
+  src="https://user-images.githubusercontent.com/813754/187436947-875dbf46-1b26-42da-b96f-e7cbe315f0eb.png"
 />
 
-## How to use color for product UI
 
-Primer delivers colors in the form of [variables](https://primer.style/primitives/colors) or design tokens. Color variables are grouped based on their purpose:
+Color design tokens are grouped based on their purpose:
 
 - **Functional:** To convey a meaning or a state. For example, from a functional perspective the color green is used to reinforce positive messaging. In a functional system, green variables are named with the suffix `.success`.
 - **Presentational:** To represent a color. For example, the color steps in the Primer scale are named by color and lightness, such as `scale.blue.5`. These variables don't currently support color modes.
--- a/content/foundations/color.mdx
+++ b/content/foundations/color.mdx
@@ -38,15 +38,25 @@ For example you use `bg-default` for the background of the page and `fg-default`
 #### Color design tokens are grouped based on their purpose:
 
 - **Presentational:** To represent a color. For example, the color steps in the Primer scale are named by color and lightness, such as `scale.blue.5`. These design tokens don't support color modes.
-  As the system grows, it will provide more APIs that fit different color needs. Check the [Primer Primitives](https://github.com/primer/primitives/releases/) repository to follow along as we release new variables and systems.
-- **Functional:** To convey a meaning or a state. For example, from a functional perspective the color green is used to reinforce positive messaging. In a functional system, green variables are named with the suffix `.success`.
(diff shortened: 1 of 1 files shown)


content/foundations/color.mdx
```diff
@@ -17,21 +17,30 @@ import {Box, Button, Heading, Label, LabelGroup, Link} from '@primer/react'
   src="https://user-images.githubusercontent.com/6951037/146927448-cb518377-114a-4ab8-a37e-1eeffc8732c7.png"
 />
 
-GitHub's UI offers a variety of different color modes. When designing product interfaces, you should design in light mode by default. Every pattern in Primer is built to work across all color modes out of the box.
+
+GitHub's UI offers a variety of different color modes. Every pattern in Primer is built to work across all color modes out of the box.
+
+When designing product interfaces in Figma, we recommend to design in light mode. This is best, because the primer components for Figma are only available in light mode.
+To preview your work in other modes, use the [Figma color mode plugin](https://www.figma.com/community/plugin/992128487074360945/Change-Color-Mode).
```
langermank, line 24:
```suggestion
When working on product interfaces in Figma, we recommend designing in light mode as our Primer Figma components are only available in light mode.
To preview your work in other modes, use the [Figma color mode plugin](https://www.figma.com/community/plugin/992128487074360945/Change-Color-Mode).
```

content/foundations/color.mdx
```diff
@@ -17,21 +17,30 @@ import {Box, Button, Heading, Label, LabelGroup, Link} from '@primer/react'
...
+
+## How to use color for product UI
+
+Primer delivers colors in the form of [design tokens](https://primer.style/primitives/colors). Design tokens are a layer of abstraction that allows better maintainability, consistancy and easy theming.
+
+For example you use `bg-default` for the background of the page and `fg-default` for the text color. If the user changes to a dark mode, we change the underlying color that those tokens reference, but the tokens stay the same.
```
langermank, line 30:
```suggestion
For example, if a page uses `bg-default` for the background color and `fg-default` for the text color, when the color mode is changed from light to dark, the tokens remain the same while their values update to respect the color mode. 
```

content/foundations/color.mdx
```diff
@@ -17,21 +17,30 @@ import {Box, Button, Heading, Label, LabelGroup, Link} from '@primer/react'
...
 
-## How to use color for product UI
-
-Primer delivers colors in the form of [variables](https://primer.style/primitives/colors) or design tokens. Color variables are grouped based on their purpose:
+#### Color design tokens are grouped based on their purpose:
 
-- **Functional:** To convey a meaning or a state. For example, from a functional perspective the color green is used to reinforce positive messaging. In a functional system, green variables are named with the suffix `.success`.
-- **Presentational:** To represent a color. For example, the color steps in the Primer scale are named by color and lightness, such as `scale.blue.5`. These variables don't currently support color modes.
+- **Presentational:** To represent a color. For example, the color steps in the Primer scale are named by color and lightness, such as `scale.blue.5`. These design tokens don't support color modes.
   As the system grows, it will provide more APIs that fit different color needs. Check the [Primer Primitives](https://github.com/primer/primitives/releases/) repository to follow along as we release new variables and systems.
```
langermank, line 41: Should this chunk remain at the end?

langermank (changes requested):
Just a few grammatical suggestions!

content/foundations/color.mdx
lukasoppermann: What are you exactly referring to?

> As the system grows, it will provide more APIs that fit different color needs. Check the [Primer Primitives](https://github.com/primer/primitives/releases/) repository to follow along as we release new variables and systems.

Or Presentational?

content/foundations/color.mdx
lukasoppermann: Hey Katie, while I love this, it moves us from a grade 9 to a grade 14 reading experience. I am trying to keep it lower. Can we break it up into multiple sentences. This always makes it easier to read.

content/foundations/color.mdx
lukasoppermann: Same as above. 😢 Love the change, but 9→14

content/foundations/color.mdx
lukasoppermann: Nevermind, I got it now and totally agree.

content/foundations/color.mdx
emilybrick: >it moves us from a grade 9 to a grade 14 reading experience
Hey @lukasoppermann, I'm a bit confused by the grade reading levels, is there somewhere that outlines what this means? 

@langermank's suggestion on line 23 feels straightforward to me - I don't know if using a conjunction makes the reading level that much more difficult (but I suppose this is subjective)

lukasoppermann: So we are suggesting here: https://primer.style/design/foundations/content#1-keep-it-clear a 7th grade or below level. Which I find very hard to reach. However using https://hemingwayapp.com/ you can get at least to a somewhat simple level.

While we are all somewhat deep in the topic colors, tokens, etc. is pretty complex and new stuff for others. Considering not everybody is an english-native speaker, I do think we should consider making out docs as easy as possible.

Please don't get me wrong, I am happy to change & update the text. But I do feel we should strive for as easy as possible.

emilybrick: Ah, I don't think I remembered this reference or didn't know how literally to take it (what defines a 7th grade reading level). That's a fair point.

I don't feel that strongly about this - if you want to keep two sentences I would simplify them both, and small fix to remove the comma in the second sentence: 

```suggestion
When designing product interfaces in Figma, we recommend using light mode. This is best because the Primer Figma components are only available in light mode.
To preview your work in other modes, use the [Figma color mode plugin](https://www.figma.com/community/plugin/992128487074360945/Change-Color-Mode).
```

content/foundations/color.mdx
lukasoppermann: This one is not so easy. I think we can't just put a fullstop in there. @langermank any idea of how to make it work with two sentences?

emilybrick:
```suggestion
For example use `bg-default` for the background of the page and `fg-default` for the text color. If the user changes to a dark mode, the underlying color that those tokens reference change, but the tokens stay the same. 
```

lukasoppermann: Love it. 👍

langermank (approved):

PR #294, opened 2022-09-14
[WIP] Single page component docs prototype

langermank:
WIP :)


src/SubNav.jsx
```diff
@@ -0,0 +1,29 @@
+import React, {useState} from 'react'
+import {UnderlineNav} from '@primer/react'
+
+export function SubNav() {
+  return (
+    <UnderlineNav
+      aria-label="Page navigation"
+      sx={{position: 'sticky', top: '66px', background: '#ffffff', borderBottom: 'none'}}
+    >
+      <UnderlineNav.Link href="#usage" selected={location.hash === '#usage'}>
```
broccolinisoup, line 10: Oh sorry I couldn't help you with this selected state! I think I was so focused on "why this component doesn't have onSelect prop" and completely missed the fact that we can always manually do it 🙈  I am glad it is all sorted out though 😊 And this looks amazing 🔥

langermank: @broccolinisoup no worries at all! If you want to use this branch to test the newer UnderlineNav when its shipped definitely feel free! 🚀

broccolinisoup: @langermank Thank you! I will 🤩

src/LookbookEmbed.jsx
```diff
@@ -0,0 +1,121 @@
...
+  const [colorScheme, setColorScheme] = useState('light')
+  const base = baseUrls[framework]
+  const [preview, setPreview] = React.useState(Object.keys(previews)[0])
+  const statusString = `${status}`
+  const componentString = `${component}`
+  const previewString = `${preview}`
+  const options = [statusString, componentString, previewString].join('/')
+  console.log(options)
+
+  const url = `${base}${new URLSearchParams(options)}`
```
langermank, line 20: JavaScript friends, can anyone tell me why `URLSearchParams` is adding a `=` to the end of the url? This seems to work to populate the iframe without the `=` 😄 

CC @jonrohan @broccolinisoup

broccolinisoup: Hi @langermank  👋🏼  I think it does it because `URLSearchParams` work with query strings and query strings are like key/value format in URLs like `baseURL?key1=value1&key2=value2` etc . See [the mdn ref](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) for examples. 

I think here the options are more like query params than query strings. (`beta/button/playground`) so I guess we don't need `URLSearchParams` here unlike we do in StorybookEmbed. We can just pass the options without creating a new `URLSearchParams` object.

broccolinisoup, line 20:
```suggestion
  const url = `${base}${options}`
```

langermank: Thank you @broccolinisoup! I didn't realize `URLSearchParams` was expecting a key/value pair 😆 just hacked my way through it. Your suggestion works perfectly! 🙌

langermank:
@emilybrick I spent the week setting us up with some infrastructure to support the single page prototype 😄 

- Lookbook and Storybook iframe components that we can pass specific stories/previews to render
- `UnderlineNav` support (its not perfect but its a start)
- `PropsTable` component that can populate from JSON data. I added some mock data for now, but it would be interesting to try and source this from both React and Rails
- Added some components and a template to this [interface guidelines Figma library](https://www.figma.com/file/YlBTqHTbdRwYQt0bfIHvGw/Interface-guidelines?node-id=38%3A27938). Might be a good place to think about how the Usage section should look (section requirements, etc.)

simurai:
FYI: https://github.com/primer/design/pull/364 includes the `StorybookEmbed` for the utilities and probably will cause merge conflicts with this PR. 😇

langermank:
Goodbye old friend

PR #296, opened 2022-09-17
Fix typo on word consitency

txomon:


lukasoppermann (approved):
Good catch.

txomon:
I don't think I can deal with that CD error

lukasoppermann:
Don't worry @txomon I will look into it and merge it. 👍

PR #298, opened 2022-09-22
Comment box

simurai:
This adds the `CommentBox` component.

👀  [Preview](https://primer-707dc87d93-26441320.drafts.github.io/components/comment-box)

A diagram of a comment box with a label, a tab nav, a toolbar and an input with a placeholder text.

---

- [Figma source](https://www.figma.com/file/BMg5gToJrKBvJCkwM09iv0/Q3---Comment-form-explorations?node-id=207%3A2321) (for the images)
- Part of https://github.com/github/primer/issues/737
- [API](https://github.com/github/primer/pull/1289)

### DIFF
--- a/content/components/comment-box.mdx
+++ b/content/components/comment-box.mdx
@@ -95,7 +95,7 @@ A label is shown above the comment box to establish the context and inform users
 
 The placeholder text is meant to give users addtional hints. Keep in mind that it will disappear as soon as users start typing. The default placeholder text is "Type here...".
 
-Note: A placeholder text is not required, but still recommended. It helps to improve usablity and make sure users know where to type.
+Note: A placeholder text is not required, but can help suggest CommentBox features specific to a component implementation, such as slash commands.
 
 <DoDontContainer>
   <Do>
--- a/content/components/comment-box.mdx
+++ b/content/components/comment-box.mdx
@@ -167,7 +167,7 @@ A [flash alert](https://primer.style/design/ui-patterns/messaging#flash-alerts) 
   width="960"
 />
 
-## Responsiveness
+## Responsive behavior
 
 The comment box's toolbar uses an [action bar](https://primer.style/design/components/action-bar) under the hood that is responsive by default. Toolbar items that don't fit in the available space are moved to an [overflow menu](https://primer.style/design/components/action-bar#overflow-menu).
 
--- a/content/components/comment-box.mdx
+++ b/content/components/comment-box.mdx
@@ -64,7 +64,7 @@ import {Box, Label, LabelGroup, Text} from '@primer/react'
 
 ## Anatomy
 
-<img width="960" alt="A diagram of a comment box with a label, a tab nav, a toolbar and an input with a placeholder text." src="https://user-images.githubusercontent.com/378023/191696249-e3fc3240-5b7f-49e5-9973-6698e705ecd6.png" />
+<img width="960" alt="A diagram of a comment box with a label, a tab nav, a toolbar and an input with a placeholder text." src="https://user-images.githubusercontent.com/378023/193514076-427273cc-4e81-4793-a401-9fc47b3c8dfb.png" />
 
 ## Content
 
@@ -75,7 +75,7 @@ A label is shown above the comment box to establish the context and inform users
 <DoDontContainer>
   <Do>
     <img
-      src="https://user-images.githubusercontent.com/378023/191459919-0301444a-18cc-474e-8db7-1480976e4128.png"
+      src="https://user-images.githubusercontent.com/378023/193514339-81bc5e18-3aab-4e15-919f-933fb5e74ef0.png"
       role="presentation"
       width="456"
     />
@@ -83,7 +83,7 @@ A label is shown above the comment box to establish the context and inform users
   </Do>
   <Dont>
     <img
-      src="https://user-images.githubusercontent.com/378023/191459943-1bc24fb4-224e-4598-94b6-f15b80401946.png"
+      src="https://user-images.githubusercontent.com/378023/193514422-54ab3bb8-238f-4852-a813-206b81449575.png"
       role="presentation"
       width="456"
     />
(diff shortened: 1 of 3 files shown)


content/components/comment-box.mdx
```diff
@@ -0,0 +1,199 @@
...
+      role="presentation"
+      width="456"
+    />
+    <Caption>Don't use the label to notify users of any kind of state or event, like when an error occurred.</Caption>
+  </Dont>
+</DoDontContainer>
+
+### Placeholder
+
+In most cases a placeholder text is not needed and the label will suffice.
```
lukasoppermann, line 96: Hey @simurai I am just curious, is this suggested by the a11y team? Personally I feel that without the placeholder it does not look so much like I could put text into it. But this may also be my personal preference.

lukasoppermann:
Very neat. 👍 

Are we removing the `Attach files by dragging ...` hint below the input?

content/components/comment-box.mdx
simurai: It's how I understood it, but might not be accurate. A proper accessibility review will still follow. I also have a hard time coming up with a good placeholder text when there is already a label above the comment box. The example in this guide says "Type here...", but that sounds a bit too blunt? 🤔

comment box show a placeholder text that reads: Type here...

But I think I know what you mean, a placeholder text is good for usability as a hint that it's an input.

simurai:
> Are we removing the `Attach files by dragging ...` hint below the input?

Yes, that's the plan. It's a bit risky because it helps new users discover that feature, but once you learned about it, it's also a bit waste of space to always show it. I was thinking that we might could surface it as part of the "pro tips" that are shown in issues/PRs, but it would not be a permanent part of the component.

Here a [CodePen](https://codepen.io/simurai/pen/QWmPXdE/43adfab66442672ddb82df6d4253e916?editors=1100) with an example that links to the [docs](https://docs.github.com/en/get-started/writing-on-github), but that could also be more detailed. Or we keep rotating pro tips randomly.

Screen Shot 2022-09-22 at 21 21 17

vdepizzol:
@simurai looking great! quick _quick_ comments:

- I think icon buttons have a more subtler color other than `fg.default`?
- We probably need to confirm the padding around the icons to make sure there's enough touch area. [Particularly in the narrow representation](https://github.com/github/design/discussions/110). 8px of padding around a 32px. Reference:
  > fine vs coarse touch target on icon buttons

- Instead of having a section called "Responsiveness", I think we can use "Responsive behavior"?

simurai:
> * I think icon buttons have a more subtler color other than `fg.default`?

Was wondering about that too for `ActionBar` in [this comment](https://www.figma.com/file/8OhcIX3PzcVhmDwSHnzwMj?node-id=97:14854#260899075).

> Right. In `.btn-octicon` we use `fg-muted`. But for the new `.Button--invisible` (that is used in ActionBar) it's [`fg-default`](https://github.com/primer/view_components/blob/36676c0f53bce4f39b858984d651c29170a4b47d/app/components/primer/beta/button.pcss#L245).

@langermank do you happen to remember if that was intentional? Possibly accessibility related? Since `.Button--invisible` shows an icon stand-alone (no border and no other text), it should use the default text color?

simurai:
> * We probably need to confirm the padding around the icons to make sure there's enough touch area. [Particularly in the narrow representation](https://github.com/github/design/discussions/110). This probably means 8px of padding around 32px-tall controls.

Is an 8px gap also needed for controls that have to border? This came up in the ActionBar API too https://github.com/github/primer/pull/1236#discussion_r966668427:

> We could, but I think having a gap between `Button--invisible` is not really needed since they have no borders and visually there is already enough spacing between the icons. With an extra gap the icons would feel too spaced out:

`gap: 0`:

Screen Shot 2022-09-09 at 15 01 04

`gap: 8px`:

Screen Shot 2022-09-09 at 15 01 33

But yes, for `pointer: coarse` there is an [extra gap](https://github.com/primer/view_components/pull/1339/files#diff-d0b676bfcc84fe135a0db9884bceafca3356b2e48d970583bb62387085185c21R38-R52) to account for the `44px` min target area.

---

> Instead of having a section called "Responsiveness", I think we can use "Responsive behavior"?

👍

content/components/comment-box.mdx
```diff
@@ -87,13 +87,15 @@ A label is shown above the comment box to establish the context and inform users
...
   </Dont>
 </DoDontContainer>
 
-### Placeholder text
+### Placeholder text (recommended)
 
-The placeholder text is meant to give users addtional hints. Keep in mind that it will disappear as soon as users start typing. The default placeholder text is "Type here..."
+The placeholder text is meant to give users addtional hints. Keep in mind that it will disappear as soon as users start typing. The default placeholder text is "Type here...".
+
+Note: A placeholder text is not required, but still recommended. It helps to improve usablity and make sure users know where to type.
```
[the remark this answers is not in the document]
simurai: Sounds good to me. 👍  The

> ...still recommended. It helps to improve usability and make sure users know where to type.

part was referring to https://github.com/primer/design/pull/298#discussion_r977339009. So even in cases where a placeholder text is not needed, it's recommended to have one just to make sure users realize that it's a `textarea` and they can type inside.

In https://github.com/github/github/pull/242242 I added a pattern where the placeholder text is just a short version of the label in case we can't come up with a better placeholder text (mention slash commands or so). E.g.

Label | Placeholder
--- | ---
Add a comment | Comment
Add a description | Description

Screen Shot 2022-10-18 at 13 16 44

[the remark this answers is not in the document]
simurai: Ok, updated the default placeholder text from `Type here...` to `Comment`. Also kept the "usability" part.

simurai:
Gonna merge this as is since we're already testing the `CommentBox` in a [prototype](https://github.com/github/primer/discussions/1313). There will likely be follow-up changes.

content/components/comment-box.mdx
```diff
@@ -93,9 +93,7 @@ A label is shown above the comment box to establish the context and inform users
 
 ### Placeholder text (recommended)
 
-The placeholder text is meant to give users addtional hints. Keep in mind that it will disappear as soon as users start typing. The default placeholder text is "Type here...".
-
-Note: A placeholder text is not required, but can help suggest CommentBox features specific to a component implementation, such as slash commands.
+The placeholder text is meant to give users addtional hints. Keep in mind that it will disappear as soon as users start typing. A placeholder text is not required, but can help suggest CommentBox features specific to a component implementation, such as slash commands. Having a placeholder text also helps with improving usablity and making sure users know where to start typing. The default placeholder text is "Comment".
```
ericwbailey, line 98: Nice!

PR #299, opened 2022-09-22
RelativeTime

josepmartins:
Tracking issue: https://github.com/github/primer/issues/1252
Figma: https://www.figma.com/file/ScxQQeqilNPokwTCaG6okn/RelativeTime-(FY23-Q1)?node-id=218%3A2711

👁️ Preview: https://primer-a7c04617fa-26441320.drafts.github.io/components/relative-time

RelativeTime examples

### DIFF
--- a/content/components/relative-time.mdx
+++ b/content/components/relative-time.mdx
@@ -11,7 +11,7 @@ import {Box, Text} from '@primer/react'
 
 ## Content
 
-### Past
+### Past dates
 
 For past dates, the relative time outputs different formats depending on the time difference.
 
--- a/content/components/relative-time.mdx
+++ b/content/components/relative-time.mdx
@@ -3,36 +3,142 @@ title: Relative Time
 status: Experimental
 ---
 
-import {Box, Label, LabelGroup, Text} from '@primer/react'
+import {Box, Text} from '@primer/react'
 
 <Box sx={{fontSize: 3}} class="lead" as="p">
   A relative time displays the relative time difference between the source date-time and the provided date-time.
 </Box>
 
+## Anatomy
+
+![A diagram of a relative time output.]()
+
 ## Content
+
+<Box
+  as="figure"
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column-reverse', 'column-reverse', 'column-reverse', 'column-reverse', 'row']}
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+<div>
 
 ### Past date-time
 
-For past date-times the relative time outputs different formats depending on the time difference:
+For past date-times the relative time outputs different formats depending on the time difference.
 
-#### Less than 1 month
+#### Within a month
 
-It shows relative date phrases in the format of `X days ago` or `X hours ago`:
+It shows relative date phrases in the format of `X seconds/minutes/days ago`. If the time difference is less than 1 minute, it shows `just now`.
 
-#### After a month has passed
+#### After a month
 
-It shows the date in the format `MMM D` or `MMM D, YYYY` depending on the current year.
+It shows the date in the format `on D MMM` or `on D MMM, YYYY` depending on the current year.
+
+</div>
+
+<Box flexShrink={0}>
+  <img
+    src="https://user-images.githubusercontent.com/912236/191943024-037847be-abc4-4242-95d4-93c3a21b2b56.png"
+    role="presentation"
+    width="464"
+  />
+</Box>
+</Box>
 
 ### Future date-time
 
-It strictly shows relative date phrases in the format of `in X days` or `in X hours`. This will never switch to displaying the date, even after a month has passed.
+It strictly shows relative date phrases in the format of `in X seconds/minutes/hours/days/months/years`. This will never switch to displaying the date, even after a month has passed.
+
+### Micro
+
+The `micro` variant shows a smaller version of the relative time. It is useful for displaying the relative time in a more compact space or in a mobile environment. It shows the relative time in the format of `Xs/h/d`, `D MMM` after a month or `D MMM, YYYY` after a year.
+
+<DoDontContainer>
+  <Do>
+    <img
+      src="https://user-images.githubusercontent.com/912236/191981619-4b2a5625-aa28-44d4-a650-fcda9f349b6f.png"
+      role="presentation"
+      width="456"
+    />
+    <Caption>Use micro to display time in a more compact space or in a mobile environment.</Caption>
+  </Do>
+  <Dont>
+    <img
+      src="https://user-images.githubusercontent.com/912236/191981619-4b2a5625-aa28-44d4-a650-fcda9f349b6f.png"
+      role="presentation"
+      width="456"
+    />
+    <Caption>Don't use the micro option when you can display time in a more prominent space.</Caption>
+  </Dont>
+</DoDontContainer>
+
+<DoDontContainer>
+  <Do>
+    <img
+      src="https://user-images.githubusercontent.com/912236/191981619-4b2a5625-aa28-44d4-a650-fcda9f349b6f.png"
+      role="presentation"
+      width="456"
+    />
+    <Caption>Use the micro option in isolation.</Caption>
+  </Do>
+  <Dont>
+    <img
+      src="https://user-images.githubusercontent.com/912236/191981619-4b2a5625-aa28-44d4-a650-fcda9f349b6f.png"
+      role="presentation"
+      width="456"
+    />
+    <Caption>Don't use it in a full sentence. Use the default format instead.</Caption>
+  </Dont>
+</DoDontContainer>
 
 ## Usage
 
-## Responsiveness
+A relative time is intended to be used within a sentence, following the action that it's relative to.
 
-There are cases where relative dates needs to be placed in a spot that is too narrow to display a long relative date. In these cases, the output can be adapted to fit in the narrow area without sacrificing usability.
+<img
+  width="960"
+  alt=""
+  src="https://user-images.githubusercontent.com/912236/191977760-bccc2197-6ecb-4487-9c5d-0d3732cc7f9e.png"
+/>
+
+### Isolated
+
+A relative time can be used in isolatation when the content and the elements sourronding make it clear what the time is relative to. For instance, when used in a list or a table where the header or titles references the action that was performed.
+
+<Box display="grid" gridTemplateColumns={['1fr', null, null, null, '1fr 1fr']} gridGap={5}>
+  <div>
+    <img
+      width="460"
+      alt=""
+      src="https://user-images.githubusercontent.com/912236/191966258-30d0b3fc-1044-4d86-aa28-9022907650a8.png"
+    />
+    <Caption>A tabe where each item title contains the action.</Caption>
+  </div>
+  <div>
+    <img
+      width="460"
+      alt=""
+      src="https://user-images.githubusercontent.com/912236/191965093-80c1858f-5d36-4caf-aecf-fd333c2c104c.png"
+    />
+    <Caption>A timeline where the title implies the main action.</Caption>
+  </div>
+</Box>
+
+### Links
+
+A relative time is used as plain text in most of the cases. However, it can be used as a link to point to a section within a view that contains relevant information.
+
+<img
+  width="960"
+  alt=""
+  src="https://user-images.githubusercontent.com/912236/191949470-a7b633fd-c3fa-48b5-a75b-e91cafdcea61.png"
+/>
+<Caption>A relative time as a link for comments in a thread.</Caption>
 
 ## Accessibility
 
-A relative time is treated like a [time](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) element when it shows a date, and it contains an [abbreviation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) element when it shows a relative date phrase.
+A relative time is treated like a [time](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) element when it shows a local numerical date, and it contains an [abbreviation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) element when it shows a relative date phrase.
--- a/content/components/relative-time.mdx
+++ b/content/components/relative-time.mdx
@@ -26,7 +26,7 @@ If the date is in past months, the date time[^2] format `on D MMM` is used. If t
 | Time difference | Format                           | Examples                                                                                   |
 | --------------- | -------------------------------- | ------------------------------------------------------------------------------------------ |
 | < 1 minute      | just now                         | `just now`                                                                                 |
-| < 1 month       | X seconds/minutes/hours/days ago | `a minute ago`, `5 minutes ago`, `one hour ago`, `3 hours ago`, `yesterday`, `20 days ago` |
+| < 1 month       | X seconds/minutes/hours/days ago | `a minute ago`, `5 minutes ago`, `an hour ago`, `3 hours ago`, `yesterday`, `20 days ago` |
 | > 1 month       | on D MMM                         | `on 18 Nov`                                                                                |
 | Past year       | on D MMM, YYYY                   | `on 26 Aug, 2021`                                                                          |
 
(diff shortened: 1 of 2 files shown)


content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+  loop
+  src="https://user-images.githubusercontent.com/912236/192255353-88749767-f1d1-49af-8145-86b60f5215b8.mp4"
+/>
+<Caption>The micro format is used for responsive layouts.</Caption>
+
+### Links
+
+A relative time is used as plain text in most cases. However, it can be used as an anchor link to point to an element within a view that contains relevant or nested information.
+
+<img
```
maximedegreve, line 106: The link on hover should have an underline.

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+      role="presentation"
+      width="456"
+    />
+    <Caption>Avoid using the micro option when there's a prominent space. Use the default option instead.</Caption>
+  </Dont>
+</DoDontContainer>
+
+## Usage
+
+A relative time is intended to be used within a sentence, following the action that it's relative to.
```
maximedegreve, line 62: I find this sentence a bit confusing because this implies it can't be used outside a sentence while in the next paragraph it can be used in isolation.

I think `A relative time can be used within a sentence` would remove this confusion.

maximedegreve:
This is great! 

I'm wondering if we could make it even better by adding a table with all the formats and examples.
When I'm designing this would save me time having to google for my use case what formats I can use.

| When | Format | Example |
|--------|--------|--------|
| 1 day + | D MMM, YYYY | 14 May, 2020 |

lukasoppermann:
Hey, in the figma file (I am assuming you meant to link to `phase 2` instead if `phase 1`) you have the following code:

```jsx
<RelativeTime micro={{narrow: true}}  datetime={Date} />
```

Why do you have to set the `micro` property to `narrow: true`? Wouldn't it be better to do `micro=true` or even better `narrow=true`?

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
+---
+title: Relative time
+status: Experimental
+---
+
+import {Box, Text} from '@primer/react'
+
+<Box sx={{fontSize: 3}} class="lead" as="p">
+  A relative time displays the relative time difference between the source date-time and the provided date-time.
```
lukasoppermann, line 9:
```suggestion
  A relative time displays the time difference between now and the specified date-time.
```

I hope this makes sense. I find the sentence a bit confusing otherwise and you repeat `relative time`.

Is the `source date-time` now or can it be a different time as well?

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+</Box>
+
+## Content
+
+### Past
```
lukasoppermann, line 14:
```suggestion
### Past dates
```

Makes it more obvious what you are talking about.

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+After a month has passed, the output shows a relative date in the format `on D MMM` or `on D MMM, YYYY` depending on the current year.
+
+<img
+  width="960"
+  src="https://user-images.githubusercontent.com/912236/192233477-8c30b28b-5f48-44c6-b373-39a482f745a1.png"
+  role="presentation"
+/>
+<Caption>A list of items using the past relative outputs.</Caption>
+
+### Future
```
lukasoppermann, line 29:
```suggestion
### Future dates
```

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+## Content
+
+### Past
+
+For past dates, the relative time outputs different formats depending on the time difference.
+
+Within the current month, the output shows a relative date in the format of `X seconds/minutes/hours/days ago`. If the time difference is less than 1 minute, it shows `just now`.
+
+After a month has passed, the output shows a relative date in the format `on D MMM` or `on D MMM, YYYY` depending on the current year.
```
lukasoppermann, line 20:
```suggestion
After a month has passed, the output shows a relative date in the format `on D MMM` if the date is within the current year or `on D MMM, YYYY` if the date is within the previous year or before.
```

I hope this makes it more clear.

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+<img
+  width="960"
+  src="https://user-images.githubusercontent.com/912236/192233477-8c30b28b-5f48-44c6-b373-39a482f745a1.png"
+  role="presentation"
+/>
+<Caption>A list of items using the past relative outputs.</Caption>
```
lukasoppermann, line 27:
```suggestion
<Caption>A list of items using relative time outputs.</Caption>
```

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### Future
+
+As well as past dates, future dates output different formats depending on the time difference.
```
lukasoppermann, line 31:
```suggestion
Future dates also use different formats depending on the time difference.
```

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+If the future date is due to happen within the current month, the relative time outputs the format of `in X seconds/minutes/hours/days`. If the time difference is less than 1 minute, it shows `just now`.
+
+If the future date is due to happen within the following months, the relative time outputs the format of `on D MMM` or `on D MMM, YYYY` when following years.
```
lukasoppermann, line 35:
```suggestion
If the date in the future but not within the current month, the format `on D MMM` is used. If the date is in the next year or further, the format `on D MMM, YYYY` is used.
```

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### Micro
+
+The `micro` variant shows a smaller version of the relative time. It is useful for displaying the relative time in a more compact space or in a mobile environment. The output shows the relative time in the format of `Xs/h/d`, `D MMM` after a month, or `D MMM, YYYY` after a year.
```
lukasoppermann, line 39:
```suggestion
The `micro` variant shows a short version of the relative time. It's useful to display the relative time in a more compact space or in a mobile environment. The output shows the relative time in the format of `Xs/h/d` or `D MMM` if the date is not within this month, or `D MMM, YYYY` if the date is not within the current year.
```

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### Future
+
+As well as past dates, future dates output different formats depending on the time difference.
+
+If the future date is due to happen within the current month, the relative time outputs the format of `in X seconds/minutes/hours/days`. If the time difference is less than 1 minute, it shows `just now`.
+
+If the future date is due to happen within the following months, the relative time outputs the format of `on D MMM` or `on D MMM, YYYY` when following years.
+
+### Micro
```
lukasoppermann, line 37:
```suggestion
### Micro format
```

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+The `micro` variant shows a smaller version of the relative time. It is useful for displaying the relative time in a more compact space or in a mobile environment. The output shows the relative time in the format of `Xs/h/d`, `D MMM` after a month, or `D MMM, YYYY` after a year.
+
+<DoDontContainer>
+  <Do>
+    <img
+      src="https://user-images.githubusercontent.com/912236/191981619-4b2a5625-aa28-44d4-a650-fcda9f349b6f.png"
+      role="presentation"
+      width="456"
+    />
+    <Caption>Use micro to display time in a more compact space or in a mobile environment.</Caption>
```
lukasoppermann, line 48:
```suggestion
    <Caption>Use the micro format to display time in a more compact space or in a mobile environment.</Caption>
```

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+  </Do>
+  <Dont>
+    <img
+      src="https://user-images.githubusercontent.com/912236/191997424-821afe8a-39d9-47c4-9203-ebaee182f3b1.png"
+      role="presentation"
+      width="456"
+    />
+    <Caption>Avoid using the micro option when there's a prominent space. Use the default option instead.</Caption>
```
lukasoppermann, line 56:
```suggestion
    <Caption>Don't use the micro format when there's a prominent space. Use the default option instead.</Caption>
```

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+<img
+  width="960"
+  alt=""
+  src="https://user-images.githubusercontent.com/912236/191977760-bccc2197-6ecb-4487-9c5d-0d3732cc7f9e.png"
+/>
+
+### Isolated
+
+A relative time can be used in isolation when the content and the surrounding elements make it clear what the time is relative to.
```
lukasoppermann, line 72:
```suggestion
A relative time can be used in isolation when the content and the surrounding elements make it clear what the time is related to.
```

I think you are referring to the related element, correct?

content/components/relative-time.mdx
```diff
@@ -0,0 +1,115 @@
...
+    />
+    <Caption>Used in a table where each item title references the action that was performed.</Caption>
+  </div>
+  <div>
+    <img
+      width="460"
+      alt=""
+      src="https://user-images.githubusercontent.com/912236/191965093-80c1858f-5d36-4caf-aecf-fd333c2c104c.png"
+    />
+    <Caption>Used in timeline where the section title references the action.</Caption>
```
lukasoppermann, line 89:
```suggestion
    <Caption>Used in a timeline where the section title references the action.</Caption>
```

content/components/relative-time.mdx
josepmartins: You are right 👍, fixed!

content/components/relative-time.mdx
josepmartins: Makes sense. The source date-time is the current time and that can't be modified. 

I'm thinking of using `current time` or `current timestamp` instead of `now`.

josepmartins: Makes sense. The source date-time is the current time and that can't be modified. 

I'm thinking of using `current time` or `current timestamp` instead of `now`.

lukasoppermann: `current time` is probably a good mix between not to technical, but still accurate.

josepmartins: rephrasing it to avoid `time` repetition

```suggestion
  A relative time displays the time difference between the current time and the specified date-time
```

content/components/relative-time.mdx
josepmartins: What about this? 

```suggestion
After a month has passed, the output shows a relative date in the format `on D MMM` if the date is within the current year or `on D MMM, YYYY` if the date is past the current year.
```

content/components/relative-time.mdx
josepmartins: I think that `If the date is in the future` might be redundant as we are in the `Future date` section. What about this?

```suggestion
If the date is in the following months, the format `on D MMM` is used. If the date is in the next year or further, the format `on D MMM, YYYY` is used.
```

content/components/relative-time.mdx
lukasoppermann: I am not 100% sure "past the current year" is correct (language wise). 🤔

content/components/relative-time.mdx
josepmartins: I rephrased it a bit, as the output is the same for past and future dates + to align with your suggestions

```suggestion
The `micro` format shows a short version of the relative time. It's useful to display the relative time in a more compact space or in a mobile environment. For both past and future dates, the output shows the relative time in the format of `Xs/h/d`. If the date is in the following or past months, the format `D MMM` is used. If the date is in the following or past years, the format `D MMM, YYYY` is used.
```

content/components/relative-time.mdx
josepmartins: good catch!

content/components/relative-time.mdx
lukasoppermann: Great 👍

content/components/relative-time.mdx
josepmartins: 🤔  what about using the structure we have in `micro` or `future` descriptions? 

Besides this, I'd follow @maximedegreve comment and add a table that simplifies and helps with these definitions.

```suggestion
If the date is in the following months, the format `on D MMM` is used. If the date is in the following years, the format `on D MMM, YYYY` is used.
```

content/components/relative-time.mdx
lukasoppermann: Perfect. 👍

josepmartins:
> Hey, in the figma file (I am assuming you meant to link to `phase 2` instead if `phase 1`) you have the following code:
> 
> ```js
> <RelativeTime micro={{narrow: true}}  datetime={Date} />
> ```
> 
> Why do you have to set the `micro` property to `narrow: true`? Wouldn't it be better to do `micro=true` or even better `narrow=true`?

Right! I changed the link 👍 . 

Following the same approach as the `hidden` option in the `SplitLayout` or the `variant` in the `SegmentControl` component, the idea is to provide a range of responsive widths (`narrow`, `regular`, `wide`) for the `micro` format, so an example of the API using micro in `narrow` spaces looks like:

```jsx
<RelativeTime micro={{narrow: true, regular: false, wide: false}} datetime={new Date("2022-12-26")}>
```

josepmartins:
> This is great!
> 
> I'm wondering if we could make it even better by adding a table with all the formats and examples. When I'm designing this would save me time having to google for my use case what formats I can use.
> 
> When	Format	Example
> 1 day >=	D MMM, YYYY	14 May, 2020

Good idea 👍 I'll add it as part of the content section.

content/components/relative-time.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+## Content
+
+### Past dates
+
+For past dates, the relative time outputs different formats depending on the time difference.
+
+Within the current month, the output shows a relative date in the format of `X seconds/minutes/hours/days ago`. If the time difference is less than 1 minute, it shows `just now`.
+
+If the date is in past months, the format `on D MMM` is used. If the date is in the past years, the format `on D MMM, YYYY` is used.
```
keithamus, line 20: Again here, it would benefit us to use [`DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) and it might be good to call it out as such.

content/components/relative-time.mdx
```diff
@@ -0,0 +1,135 @@
...
+/>
+<Caption>A list of items using relative time outputs.</Caption>
+
+### Future dates
+
+Future dates also use different formats depending on the time difference.
+
+If the future date is due to happen within the current month, the relative time outputs the format of `in X seconds/minutes/hours/days`. If the time difference is less than 1 minute, it shows `just now`.
+
+If the date is in the following months, the format `on D MMM` is used. If the date is in the next year or further, the format `on D MMM, YYYY` is used.
```
keithamus, line 42: Same for these as above, calling out the underlying API may help folk.

keithamus (approved):

lukasoppermann (approved):

content/components/relative-time.mdx
```diff
@@ -0,0 +1,140 @@
...
+Future dates also use different formats depending on the time difference.
+
+If the future date is due to happen within the a month's time, the relative time outputs the format of `in X seconds/minutes/hours/days`. If the time difference is less than 1 minute, it shows `just now`.
+
+If the date is in the following months, the format `on D MMM` is used. If the date is in the next year or further, the format `on D MMM, YYYY` is used.
+
+| Time difference | Format                                 | Examples                                                                            |
+| --------------- | -------------------------------------- | ----------------------------------------------------------------------------------- |
+| < 1 minute      | just now                               | `just now`                                                                          |
+| < 1 month       | X seconds / minutes / hours / days ago | `in a minute`, `in 2 minutes`, `in an hour`, `in 7 hours`, `tomorrow`, `in 14 days` |
```
simurai, line 52:
```suggestion
| < 1 month       | in X seconds / minutes / hours / days | `in a minute`, `in 2 minutes`, `in an hour`, `in 7 hours`, `tomorrow`, `in 14 days` |
```

PR #301, opened 2022-09-23
Add Foundations/Getting started page

vdepizzol:
This PR adds a "Getting started" page to the Foundations section, as an introduction to the Layout docs (and other foundation-related materials) that are being worked on right now.

The content of this introduction to foundations is based on the presentation "The State of UI at GitHub", which is available for hubbers in https://github.com/github/design/discussions/110.

### DIFF
--- a/content/foundations/getting-started.mdx
+++ b/content/foundations/getting-started.mdx
@@ -21,7 +21,7 @@ Primer focuses on the user experience so people can benefit from using a platfor
 
 GitHub is a home for all developers. To be inclusive means we must consider accessibility at the core of how we design.
 
-Accessibility and inclusive design needs to be expected from the start of a project, because it deeply affects its design, how the product behaves, and how it’s built.
+Accessibility and inclusive design needs to be considered from the start of a project, because it deeply affects its design, how the product behaves, and how it’s built.
 
 Adding accessibility support to a design after the fact generally turns into a difficult retrofit. To make GitHub fully available for everyone, accessibility requirements need to be taken into account during ideation, design, and implementation phases.
 
--- a/content/foundations/getting-started.mdx
+++ b/content/foundations/getting-started.mdx
@@ -1,0 +1,74 @@
+---
+title: Getting started
+---
+
+import {Box} from '@primer/react'
+
+
+<Box sx={{fontSize: 3}} class="lead" as="p">
+  Primer is built for GitHub by GitHub. Its guidelines are a collection of open source principles, standards, and recommendations for designing GitHub. As a design language, Primer provides common grammar and vocabulary to build cohesive, accessible, responsive, and efficient experiences.
+</Box>
+
+## A cohesive experience
+
+GitHub has a global audience. Millions of people use GitHub every day for many different reasons, in many different ways. They all come from different backgrounds. We can’t infer their physical or cognitive abilities, or devices they use, or the internet connection they have. All of that needs to be taken into account before, during, and after the design process.
+
+Familiar patterns help people intuitively navigate to the information they're looking for, while also helping people discover new features. If pages and features don’t share the same metaphors, people get confused more easily, it makes it harder for them to understand the features we provide.
+
+Primer focuses on the user experience so people can benefit from using a platform that is cohesive, no matter where in the platform they are. If a page behaves and feels familiar, it allows people to focus on the task at hand, instead of requiring them to parse UI that looks completely new.
+
+## Inclusive and accessible design
+
+GitHub is a home for all developers. To be inclusive means we must consider accessibility at the core of how we design.
+
+Accessibility and inclusive design needs to be expected from the start of a project, because it deeply affects its design, how the product behaves, and how it’s built.
+
+Adding accessibility support to a design after the fact generally turns into a difficult retrofit. To make GitHub fully available for everyone, accessibility requirements need to be taken into account during ideation, design, and implementation phases.
+
+For more information, see [Accessibility at GitHub](/accessibility/accessibility-at-github).
+
+## Responsive design
+
+Browsers are found in devices with all different types of shapes and sizes, and we can’t assume a desktop-only experience is enough. Supporting responsive experiences is an essential part of developing for the Web today. In fact, making sure pages work well [on most screen sizes or zoom levels](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) is an accessibility requirement as well.
+
+Nielsen Norman Group [defines Responsive Web Design](https://www.nngroup.com/articles/responsive-web-design-definition/) as “a web development approach that creates dynamic changes to the appearance of a website, depending on the screen size and orientation of the device being used to view it”.
+
+“Adaptive design”, on the other hand, has been a term historically used to identify an approach where entirely separated fixed layouts were created to support mobile devices (MDN has good documentation [on how we got here](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Responsive_Design#historic_website_layouts)).
+
+“Responsive design” at GitHub is inherently adaptive, in the sense that it should not only adjust layout and spacing when resizing a page, but support patterns and component variants that match the paradigms and affordances of the person’s device:
+
+- Responsive to the form factor: adapt to viewport size, and pointing device support (`fine` for desktop, or `coarse`, for touch), and to the device metaphors and affordances.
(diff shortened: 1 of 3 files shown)


keithamus (approved):
LGTM 👍

joshblack (approved):

content/foundations/getting-started.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+<Box sx={{fontSize: 3}} class="lead" as="p">
+  Primer is built for GitHub by GitHub. Its guidelines are a collection of open source principles, standards, and recommendations for designing GitHub. As a design language, Primer provides common grammar and vocabulary to build cohesive, accessible, responsive, and efficient experiences.
+</Box>
+
+## A cohesive experience
+
+GitHub has a global audience. Millions of people use GitHub every day for many different reasons, in many different ways. They all come from different backgrounds. We can’t infer their physical or cognitive abilities, or devices they use, or the internet connection they have. All of that needs to be taken into account before, during, and after the design process.
+
+Familiar patterns help people intuitively navigate to the information they're looking for, while also helping people discover new features. If pages and features don’t share the same metaphors, people get confused more easily, it makes it harder for them to understand the features we provide.
```
lukasoppermann, line 16: Can we invert this and formulate the positive example? "If pages and features share the same metaphors..."

content/foundations/getting-started.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+Primer focuses on the user experience so people can benefit from using a platform that is cohesive, no matter where in the platform they are. If a page behaves and feels familiar, it allows people to focus on the task at hand, instead of requiring them to parse UI that looks completely new.
+
+## Inclusive and accessible design
+
+GitHub is a home for all developers. To be inclusive means we must consider accessibility at the core of how we design.
+
+Accessibility and inclusive design needs to be expected from the start of a project, because it deeply affects its design, how the product behaves, and how it’s built.
```
lukasoppermann, line 24:
```suggestion
Accessibility and inclusive design needs to be considered from the start of a project, because it deeply affects its design, how the product behaves, and how it’s built.
```

content/foundations/getting-started.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+Adding accessibility support to a design after the fact generally turns into a difficult retrofit. To make GitHub fully available for everyone, accessibility requirements need to be taken into account during ideation, design, and implementation phases.
```
lukasoppermann, line 26:
```suggestion
Adding accessibility support to a design after the fact generally turns into a difficult retrofit.
```

I have the feeling this was already said in the previous paragraph.

content/foundations/getting-started.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+Primer patterns and components aim towards accessible solutions that don’t dilute efficient UI interactions. It focuses on responsive design that doesn’t dilute desktop experiences into mobile-first solutions. Ultimately, it recognizes the need to encourage flow by providing calm and logical experiences throughout.
+
+## The zen of GitHub
+
+The [GitHub API](https://api.github.com/zen) consolidates the process and style in which GitHub makes design decisions in 14 aphorisms.
+
+More details about the zen of GitHub can be found in [@kneath](https://github.com/kneath)’s [post](https://warpspire.com/posts/taste).
+
+- Responsive is better than fast
```
lukasoppermann, line 61: I somehow feel it would be nice to add a bit more context to those. Can't we add a "one sentence why" to every one? Or what is the effect of something e.g. `Encourage flow`. Otherwise it feels a bit excluding and top-down.

lukasoppermann:
Can we add a visual to the page? Otherwise it is a lot of text which may discourage a lot of people from reading it. At least an intro picture, maybe some octocats that show diversity?

content/foundations/getting-started.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+Accessibility and inclusive design needs to be expected from the start of a project, because it deeply affects its design, how the product behaves, and how it’s built.
+
+Adding accessibility support to a design after the fact generally turns into a difficult retrofit. To make GitHub fully available for everyone, accessibility requirements need to be taken into account during ideation, design, and implementation phases.
+
+For more information, see [Accessibility at GitHub](/accessibility/accessibility-at-github).
+
+## Responsive design
+
+Browsers are found in devices with all different types of shapes and sizes, and we can’t assume a desktop-only experience is enough. Supporting responsive experiences is an essential part of developing for the Web today. In fact, making sure pages work well [on most screen sizes or zoom levels](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) is an accessibility requirement as well.
```
lukasoppermann, line 32:
```suggestion
Browsers are found on devices with different shapes and sizes, so we can’t assume a desktop-only experience is enough. Supporting responsive experiences is an essential part of developing for the Web today. In fact, making sure pages work well [on most screen sizes or zoom levels](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) is an accessibility requirement as well.
```

content/foundations/getting-started.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+Nielsen Norman Group [defines Responsive Web Design](https://www.nngroup.com/articles/responsive-web-design-definition/) as “a web development approach that creates dynamic changes to the appearance of a website, depending on the screen size and orientation of the device being used to view it”.
```
lukasoppermann, line 34: Could we turn this into a quote block to add a little visual interest and break up the text?

content/foundations/getting-started.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+“Adaptive design”, on the other hand, has been a term historically used to identify an approach where entirely separated fixed layouts were created to support mobile devices (MDN has good documentation [on how we got here](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Responsive_Design#historic_website_layouts)).
```
lukasoppermann, line 36: Is this important or could we maybe remove this paragraph to shorten the text a bit?

content/foundations/getting-started.mdx
```diff
@@ -0,0 +1,74 @@
+---
+title: Getting started
```
lukasoppermann, line 2: I would have expected a "setup & install" post with this title. Could we maybe change it to something like "Introduction" or similar?

content/foundations/getting-started.mdx
vdepizzol: @lukasoppermann I think this is an important consideration to clear any possible confusion about our "responsive design" approach being "adaptive". I agree it might be a little confusing right now when thinking about the granularity of the other information — in the future we may move this to a more specific "Responsive design" dedicated page.

For now I think we can treat this information structure iteratively :v:

content/foundations/getting-started.mdx
vdepizzol: I'm not sure I want to highlight this paragraph in this page in particular at the moment. Once we have a dedicated "Responsive design" page I think it's worth considering it. 😁

content/foundations/getting-started.mdx
vdepizzol: @lukasoppermann the following paragraph already covers that:
> If a page behaves and feels familiar, it allows people to focus on the task at hand, instead of requiring them to parse UI that looks completely new.

I think the repetitiveness in this case is coming from an attempt to make this as clear as possible. What do you think?

content/foundations/getting-started.mdx
vdepizzol: That's a great idea. I think we should have "The Zen of GitHub" as a dedicated page in the future.

content/foundations/getting-started.mdx
vdepizzol: I think a "setup & install" makes sense for the "Build" section, but not for "Interface guidelines" in this case. I'm treating the information architecture and how we break these sections apart as a different pull request, we can improve this out iteratively!

For reference @emilybrick has been thinking about IA [in this Google Doc](https://docs.google.com/document/d/1YcIp34rA8jC7df2Vm5HUCCd1k2O_EMsdI2eO-vMiSb0/edit#heading=h.b161blt5ckx8) (only available for hubbers)

vdepizzol:
@lukasoppermann thank you so much for the feedback! I applied some of your suggestions. I know not all conversations are marked as solved, I think we can handle most of your considerations iteratively in future PRs. 

🎉 🎉 🎉

PR #319, opened 2022-11-02
Use indented do/don't instead of images in Foundations/content page

vdepizzol:
I've done a quick translation from image-based content references to an indented [Do/Dont doctocat component](https://primer.style/doctocat/components/do-dont#indented).

This should make it easier to maintain this page as we continue to evolve this documentation.

Also, just to emphasize, all of the alternative text from the now removed images are still available in the page content.


jonrohan (approved):
Looks great 👍🏻

josepmartins (approved):
🚀 ✨

PR #331, opened 2022-11-24
add more info to relative-time guidelines

keithamus:
This adds some more information to RelativeTime guidelines. It notes that users should reach for the component whenever displaying any datetime (as the component will display dates localised to the users locale), and also expands some information about when to use different date displays.

The title sections are organised as follows now:

 - Overview
   - Past Dates
   - Future Dates
   - Micro Format
   - Elapsed or remaining dates
   - Precise dates
 - Usage
 - Accessibility


content/components/relative-time.mdx
```diff
@@ -1,16 +1,24 @@
 ---
-title: Relative time
+title: Dates & Times
```
keithamus, line 2: I'm tempted to rename this so it is clear that we expect all dates & times to use this format side wide. Thoughts?

content/components/relative-time.mdx
```diff
@@ -1,16 +1,24 @@
...
 status: Experimental
 description: A relative time displays the time difference between the current time and the specified date-time.
 ---
 
 import {Box} from '@primer/react'
 
-## Content
+## Overview
+
+All dates and times displayed should use the Relaive Time component, which is able to format and localise dates into the users given locale, as well display times relative to the current date, while also updating in real time.
```
josepmartins, line 11: suggestions according to this ADR on component naming: https://github.com/github/primer/blob/main/adrs/2022-02-15-spelling-of-component-names.md

```suggestion
All dates and times displayed should use the relative time component, which is able to format and localise dates into the users given locale, as well as display times relative to the current date, while also updating in real-time.
```

content/components/relative-time.mdx
josepmartins: I think it makes sense with the new `format` approach. My only concern if we do this is the parity with the `RelativeTime` component name.

content/components/relative-time.mdx
```diff
@@ -1,16 +1,24 @@
...
+
+Dates and times almost always appear relative to the current date, using short-hand formats to help users quickly determine how old something is or how soon something will be. For this we use Relative Time which displays dates rounded to a nearest unit, for example "a minute ago" or "20 days ago". For dates that are more than 1 month from the current date it can be more useful to display the date-time, as these items are less "current" and more "historical".
```
josepmartins, line 13: This is great ❤️!  some nit-pick suggestions: 

```suggestion
Dates and times almost always appear relative to the current date, using short-hand formats to help users quickly determine how old something is or how soon something will be. For this, we use Relative Time which displays dates rounded to the nearest unit, for example, "a minute ago" or "20 days ago". For dates that are more than 1 month from the current date, it can be more useful to display the date-time, as these items are less "current" and more "historical".
```

content/components/relative-time.mdx
```diff
@@ -1,16 +1,24 @@
...
+
+Some dates and times represent the start or (expected) end of an actively running task, for these the elapsed or remaining time should be displayed, as this information is more useful than when a task started. This format should display all units that are greater than zero, for example "2 minutes, 24 seconds" or "8 hours, 15 seconds". For ephemeral tasks where timing needs to be quickly consumed, or where a lot of times may be displayed together, a compact time notation can be used, for example "2m 24s" or "8h 15s".
```
josepmartins, line 15:
```suggestion
Some dates and times represent the start or (expected) end of an actively running task, for these the elapsed or remaining time should be displayed, as this information is more useful than when a task started. This format should display all units that are greater than zero, for example, "2 minutes, 24 seconds" or "8 hours, 15 seconds". For ephemeral tasks where timing needs to be quickly consumed, or where a lot of times may be displayed together, a compact time notation can be used, for example, "2m 24s" or "8h 15s".
```

content/components/relative-time.mdx
```diff
@@ -1,16 +1,24 @@
...
+
+Some dates are especially sensitive and always should be displayed in a precise date-time format, such as the creation or expiration of certificates and keys. When a user needs to take action before a particular time (such as an expiry date) then the precise date time should be displayed. These should display the weekday, the date, the month and the year (if the year is not the current year), for example "Thursday, 26 August 2021" or "Saturday, 31 December 2022".
```
josepmartins, line 17:
```suggestion
Some dates are especially sensitive and always should be displayed in a precise date-time format, such as the creation or expiration of certificates and keys. When a user needs to take action before a particular time (such as an expiry date) then the precise date time should be displayed. These should display the weekday, the date, the month, and the year (if the year is not the current year), for example, "Thursday, 26 August 2021" or "Saturday, 31 December 2022".
```

content/components/relative-time.mdx
```diff
@@ -52,7 +60,7 @@ If the date is in the following months, the format `on D MMM` is used. If the da
 
 ### Micro format
 
-The `micro` format shows a short version of the relative time. It's useful to display the relative time in a more compact space or in a mobile environment. For both past and future dates, the output shows the relative time in the format of `Xs/h/d`. If the date is in the following or past months, the format `D MMM` is used. If the date is in the following or past years, the format `D MMM, YYYY` is used.
+The `micro` format shows a shorter version of the relative time. This can be used in cases where it's useful to display the relative time in a more compact space or in a mobile environment. For both past and future dates, the output shows the relative time in the format of `Xs/h/d`. If the date is in the following or past months, the format `D MMM` is used. If the date is in the following or past years, the format `D MMM, YYYY` is used.
```
josepmartins, line 63: I'll refactor this part on a follow-up PR 👍

content/components/relative-time.mdx
```diff
@@ -1,16 +1,24 @@
...
+
+Some dates are especially sensitive and always should be displayed in a precise date-time format, such as the creation or expiration of certificates and keys. When a user needs to take action before a particular time (such as an expiry date) then the precise date time should be displayed. These should display the weekday, the date, the month and the year (if the year is not the current year), for example "Thursday, 26 August 2021" or "Saturday, 31 December 2022".
 
 ### Past dates
 
-For past dates, the relative time outputs different formats depending on the time difference.
+For past dates, the Relative Time component outputs different formats depending on the time difference.
```
josepmartins, line 21:
```suggestion
For past dates, the relative time component outputs different formats depending on the time difference.
```

josepmartins (approved):
This is great @keithamus! I love the new `format` API approach ✨ 

Leaving some grammar/nit-picking suggestions, other than that LGTM! I'll create a follow-up PR to cover the responsive approach to relative dates instead of the `micro` option, as we discussed 👍 .

PR #333, opened 2022-11-30
Remove truncation docs

alexislucio:
Will republish at a later time.


ericwbailey (approved):
@alexislucio and I discussed this change via a Zoom chat.

andrialexandrou:
For posterity, could you say why?

PR #336, opened 2022-12-02
DataTable interface guidelines

mperrotti:


content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+
+<div>
+
+If page length is too short, it would be annoying to browse the data in small segments. If page length is too long, users could get lost or overwhelmed.
+
+The following factors influence what the ideal page length would be for your use-case:
+
+- Whether rows can be filtered
+- Number of columns
+- Visual complexity of the rendered cells
```
ericwbailey, line 156:
```suggestion
- Visual complexity of the rendered cells
- Amount of context needed when reviewing the data
```

content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+    />
+    <Caption role="presentation" color="fg.subtle">
+      Don't repeat or rephrase the title in the page header as the table title
+    </Caption>
+  </Dont>
+</DoDontContainer>
+
+## Options
+
+### Column widths
```
ericwbailey, line 276: Do we explicitly say anywhere that tables grow to fit 100% of the width of the container they are placed in?

content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+If all columns have width constraints set and the row doesn't fill the table container, the difference is added to the width of the last column.
+
+### Cells
+
+<Box display="flex" flexDirection="column" sx={{gap: 3, 'h4': { margin: 0 }}}>
+<Box display="flex" flexDirection={['column', 'column', 'column', 'column', 'row']} sx={{gap: 3}}>
+
+<Box flex={1}>
+
+#### Density
```
ericwbailey, line 363: Should we provide examples of when you'd want to adjust the default density? Or avoid adjusting it?

content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+  src="https://user-images.githubusercontent.com/2313998/205361229-8f29b673-93d5-414a-8e77-aa7b74e3c4b1.png"
+/>
+
+</Box>
+
+<Box display="flex" flexDirection={['column', 'column', 'column', 'column', 'row']} sx={{gap: 3}}>
+
+<Box flex={1}>
+
+#### Text alignment
```
ericwbailey, line 383: Similarly, should we provide examples of when and when not to adjust default alignment?

content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+  role="presentation"
+  src="https://user-images.githubusercontent.com/2313998/205361247-dfea530d-9989-442b-8a50-9c66aa5d3733.png"
+/>
+</Box>
+
+<Box display="flex" flexDirection={['column', 'column', 'column', 'column', 'row']} sx={{gap: 3}}>
+
+<Box as="p" flex={1}>
+  To handle cases where the row header can be very long, we should give consumers the option to specify a shorter string
+  to identify the rows. For example, if an issue title is used as the row header, we could use the issue number instead
```
ericwbailey, line 577: I'll update the API doc to reflect this.

content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+/>
+</Box>
+</Box>
+
+### Keyboard navigation
+
+TBD - I need help from an a11y designer
+
+<!-- TODO: get guidance from Eric -->
+
```
ericwbailey, line 594: I'm thinking we remove this section—I'm not aware of browser-specific default keyboard behavior for tables. The keyboard interactivity we'd be expecting will come from: 

- Native elements for sorting/action buttons
- Links in cell content
- Subcomponents such as ActionBar

In previous work, James has advised not listing out keyboard behavior for native controls (`textarea` behavior for the CommentBox), so this is coming from that.

content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+- **Title:** A concise label that describes the table and its contents
+- **Subtitle:** A brief description to adds more context for the table when the title alone is not enough
+- **Actions:** Controls that affect the entire table
+- **Filter:** An input used to filter table data to make it easier for the user to focus only on the rows relevant to their task
+
+### Column headers
+
+A very short label that describes the data in the column.
+
+Column headers are buttons that sort rows by the column data in ascending or descending order. Sort functionality may be disabled.
```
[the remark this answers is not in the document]
mperrotti: That's the default behavior, but the sort behavior can be disabled on a column-by-column basis.

content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+### Pagination footer
+
+The pagination footer shows the user where they are in relation to the total number of rows and provides controls for navigating all rows.
+
+## Usage
+
+### When to use a data table
+
+- To navigate a lot of information at once and to make it easy to compare data points between rows or columns
+- If the data is easy to understand when all data points are displayed in a flat hierarchy
```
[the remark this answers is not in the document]
mperrotti: I'm having a hard time understand what you mean by this

content/components/data-table.mdx
mperrotti: Yea that makes sense to me. I couldn't think of any browser-specific default keyboard behavior for tables either, and I just assumed I didn't know any better.

content/components/data-table.mdx
```diff
@@ -0,0 +1,598 @@
...
+/>
+
+### Table header
+
+The table header shows contextual information about the table and render controls that affect the table.
+
+- **Title:** A concise label that describes the table and its contents
+- **Subtitle:** A brief description to adds more context for the table when the title alone is not enough
+- **Actions:** Controls that affect the entire table
+- **Filter:** An input used to filter table data to make it easier for the user to focus only on the rows relevant to their task
```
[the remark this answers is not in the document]
mperrotti: I took a very simple approach for this by just saying exactly what you said in the next sentence.

content/components/data-table.mdx
mperrotti: No, but do you think we need to?

content/components/data-table.mdx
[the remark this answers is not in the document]
emilybrick: I'm also a bit confused by this suggestion ^

content/components/data-table.mdx
```diff
@@ -0,0 +1,604 @@
...
+  width="456"
+  role="presentation"
+  src="https://user-images.githubusercontent.com/2313998/205361249-763772e1-d856-479e-a632-d185c27095db.png"
+/>
+</Box>
+</Box>
+
+## Related components and patterns
+
+- [ActionBar](https://primer.style/design/components/action-bar)
```
emilybrick, line 603: sorry, this is probably a silly question, but why is ActionBar in here?

emilybrick (approved):
This is awesome Mike. Thank you for writing these up 🙌 . I left one comment re: related components. 

Also, in the future we may want to revise our anatomy graphic guidance. 

It's a little hard to distinguish some of the elements being called out in the Anatomy graphic, e.g. the dot overlapping the border and the dotted line. This seems like more of a symptom of how complex this particular component is (especially to view on my teeny laptop). 
Screen Shot 2022-12-05 at 9 43 14 PM

content/components/data-table.mdx
ericwbailey: I think it'd be worth it, but I'm also paranoid about these sorts of things.

content/components/data-table.mdx
mperrotti: It's used for the row actions

content/components/data-table.mdx
```diff
@@ -0,0 +1,604 @@
...
+<Box display="flex" flexDirection={['column', 'column', 'column', 'column', 'row']} sx={{gap: 3}}>
+
+<Box as="p" flex={1}>
+  The "query builder" pattern is used to show rows that match the user's query.
+</Box>
+
+<img
+  width="456"
+  role="presentation"
+  src="https://user-images.githubusercontent.com/2313998/205361241-318146a6-d2ff-47e7-96fc-4f1bd4cd5698.png"
```
joshblack, line 530: I wasn't sure what the query builder pattern that is described looks like in this image. Based on the anatomy image, is this intended to show the filter present in the table header or is there something else that I should be noting in this image? 👀

content/components/data-table.mdx
```diff
@@ -0,0 +1,604 @@
...
+</Box>
+
+<img
+  width="456"
+  role="presentation"
+  src="https://user-images.githubusercontent.com/2313998/205361244-8a1f527a-e1f3-45a3-ab13-5fdf82476038.png"
+/>
+</Box>
+
+### Labeling actions
```
joshblack, line 568: Quick question, will the labels described in this section also show up as tooltips when focusing or hovering over the icon button?

content/components/data-table.mdx
```diff
@@ -0,0 +1,604 @@
...
+
+<Box display="flex" flexDirection="column" sx={{gap: 3}}>
+<Box display="flex" flexDirection={['column', 'column', 'column', 'column', 'row']} sx={{gap: 3}}>
+
+<Box as="p" flex={1}>
+  The row header string will be prepended to the ARIA text of action buttons. For examples, Download {`{row header}`}“,
+  “Actions:{`{row header}`}”
+</Box>
```
joshblack, line 576: This seems like it could add extra verbosity that a screen reader user might not expect and couldn't control with their verbosity settings. Something that could be confusing when navigating between rows with this labeling technique is that the row header could end up being announced twice.

If I understand right, navigating tables with NVDA/JAWS/VoiceOver works like:

- If it supports announcing a cell, it will read both the column and row headers before the cell content
- When changing columns, it will announce the column header before the cell content
- When changing rows, it will announce the row header (if it exists)

Tried to confirm this over in a [Loom video](https://www.loom.com/share/6d7996881cfb45d9b7aa320c9104f2fe). Let me know if I tested anything incorrectly!

Would it make sense for the row action to have its semantic label and rely on the markup semantics of the row header to add the context for the end user? Would love to hear your thoughts here

joshblack (commented):
Loving this! I feel like it's coming together so well 🥳 

Left two questions around labeling actions, would love to hear what you all think 👀

content/components/data-table.mdx
mperrotti: Oops - I somehow accidentally deleted the filter input 🙈

I'll add it back in

content/components/data-table.mdx
mperrotti: I wasn't planning on that, but we could do that

content/components/data-table.mdx
mperrotti: Yea, I think it makes sense to let the column header and row header semantics add context for the end user.

cc @ericwbailey - what do you think of that approach?

content/components/data-table.mdx
mperrotti: I also added a link to the documentation that describes filter inputs

content/components/data-table.mdx
joshblack: @mperrotti it seems like this could fall under the guidelines of: if it's a recognizable icon, you might be okay (like close for a dialog) but otherwise you should include a tooltip with the label/text alternative for the icon button. Curious what you think!

mperrotti: Yup, I think that's the way to go 👍

mperrotti:
@ericwbailey and @joshblack - do you feel ok about merging this?

mperrotti:
I'm going to merge this by the end of today unless there are any objections

content/components/data-table.mdx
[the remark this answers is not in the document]
ericwbailey: Closing, as I think this suggestion harms more than it helps.

joshblack (approved):
Looks good to me! Only thing left was the row action semantics in: https://github.com/primer/design/pull/336/files#r1043764096, excited to see what Eric shares 👀

content/components/data-table.mdx
ericwbailey: I like it! 

My one concern is this might turn it into a content authoring issue that we can't anticipate, but I don't think that should block this progress.

Also thank you for the Loom video, @joshblack! That was super helpful 🎉

ericwbailey (approved):
Love it, merge it 🎉

PR #342, opened 2022-12-19
Fix date and times link

josepmartins:
- [x] Fixed broken link to Dates and Times patterns (formerly Relative Time)
- [x] Removed empty `textInputField` file
- [x] Removed status in components other than `Experimental` as there is no way to mark PRC and PVC status separately. Working on a draft layout in Doctocat to fix that. See preview:  https://primer-7c6b3006f4-26493675.drafts.github.io/components/test

### DIFF
--- a/content/components/action-list.mdx
+++ b/content/components/action-list.mdx
@@ -1,6 +1,5 @@
 ---
 title: Action list
-status: Alpha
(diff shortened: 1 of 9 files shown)


lukasoppermann (approved):
I am trusting you on the status.

PR #344, opened 2022-12-21
Adjust dates and times guidelines

josepmartins:
Improve dates and times guidelines according to the latest changes of the [relative-time-element](https://github.com/github/relative-time-element), aligning to the three main behaviors that we use to format dates: `relative`, `precise`, and `elapsed/duration`.

Among others, this change affects `micro` usage. This format causes accessibility issues with assistive technologies and native browser translation, and we recommend using a relative, long format instead.

Screenshot 2022-12-22 at 14 20 59

More info: https://github.com/github/relative-time-element/issues/220 

cc: @keithamus

### DIFF
--- a/content/ui-patterns/dates-and-times.mdx
+++ b/content/ui-patterns/dates-and-times.mdx
@@ -59,7 +59,6 @@ The relative dates can be used within a sentence, following the action that it's
 <Box display="grid" gridTemplateColumns={['1fr', null, null, null, '1fr 1fr']} gridGap={5}>
   <div>
     <img
-      width="460"
       alt=""
       src="https://user-images.githubusercontent.com/912236/208956681-5129e245-2752-4efb-9605-9fd33d762ceb.png"
     />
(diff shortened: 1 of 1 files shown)


keithamus (approved):
I think this is great!

As a follow up it might be nice to revisit the elapsed running times. Playing around with VoiceOver, it seems to read `03:30` as "3 minutes, 30 seconds", while `3m 30s` is "3 meters 30 s". This leads me to wonder, is the digital readout of `03:30` better for elapsed times like actions runners?

josepmartins:
> I think this is great!
> 
> As a follow up it might be nice to revisit the elapsed running times. Playing around with VoiceOver, it seems to read `03:30` as "3 minutes, 30 seconds", while `3m 30s` is "3 meters 30 s". This leads me to wonder, is the digital readout of `03:30` better for elapsed times like actions runners?

👍 I agree 100%. I had similar thoughts when writing the `micro` description and its accessibility issues. As an example, Material design follows the `H:MM:SS` format for the duration as well, and I think it makes a lot of sense for our use cases. 

https://m2.material.io/design/communication/data-formats.html#date-and-time

PR #359, opened 2023-01-18
add section about demarcating links in text

lukasoppermann:
I added a bit more details about accessibility for links.

I want to add some more graphical examples as well, but would like @ericwbailey to check this first.

Am I missing anything? Is something wrong?


ericwbailey (approved):
Looks good to me! There might also be an opportunity here to link to our newly-merged [link and button guidance](https://github.com/github/accessibility-playbook/blob/main/content/guidance/patterns/link-and-button-guidance.mdx) document.

PR #360, opened 2023-01-23
adding images and updated designer tips

lukasoppermann:
This PR adds new images to the design do's and don'ts section. It also slightly modifies the text of the section.

I already asked @ericwbailey to check the provided options. I know that not everyone favors all options, but I think if they are legit we should put them in to provide all possibilities for our designers.

Maybe we can add another paragraph somewhere to say which options are preferred and which ones should only be used if nothing else works.


ericwbailey (approved):
Apologies this took me some time to get to. I love this work, and it's a welcome addition.

ichelsea (approved):
+1 to what @ericwbailey said.

PR #363, opened 2023-01-24
fixing broken links

### LINKED ISSUE #362: Broken links in interface guidelines
I was reading the [Primer interface guidelines for icons](https://primer.style/design/components/icon) and noticed that quite a few of the links are broken. For example, the **Library** section contains a link:

> At Primer, we utilize the [Octicons](https://primer.style/design/octicons/) icon library.

Looking at the [source](https://github.com/primer/design/blob/960b9cedb7bc3a30a9b4ad2bb7dfe16e407a88f3/content/components/icon.mdx?plain=1#L19) for this link, it's obviously meant to point to `https://primer.style/octicons/`, which is the correct URL, but something on the site is prepending `/design` to the path, even though the link starts with a slash. This means the document links to `https://primer.style/design/octicons/` instead, which is 404'ing.

This problem also exists for many other links on the same page, and possibly other pages in the guidelines.

lukasoppermann:
This should fix #362 for https://primer.style/design/components/icon.

It also fixes a broken link on the homepage to the new figma section


josepmartins (approved):
Changes look good to me, but bear in mind we'll need to change them again to relative paths with the upcoming Octicons docs migration. 

Maybe this is something we could add in the task list here cc @emilybrick 
 @colebemis  https://github.com/github/primer/issues/1630

PR #365, opened 2023-01-25
Copy content from primer.style/desktop → primer.style/design

broccolinisoup:
Closes https://github.com/github/primer/issues/1615

Questions :) 

- I added index.mdx for `/native` and `/desktop` but I am not sure if we need them? 🤔 
- Do we need ToC as we have the sections on the right hand side? https://primer-3ee5eb8adb-26441320.drafts.github.io/native/desktop/foundations
- I am not sure if [the PR reference on the spacing section](https://primer-3ee5eb8adb-26441320.drafts.github.io/native/desktop/foundations#spacing) is still helpful. Would it be helpful to check the dateness of the content with the desktop team maybe?


broccolinisoup:
@emilybrick it is giving me an error because I used the same title but different sections. https://github.com/DavidAnson/markdownlint/blob/v0.26.2/doc/Rules.md#md024

I don't know how to pass parameters to the markdown so that it allows the same heading in the siblings 🤔

emilybrick:
Hey! Answering some questions:

> I added index.mdx for /native and /desktop but I am not sure if we need them? 🤔

The only content you need on the index.mdx for /desktop are links to Getting Started + Foundations. Similar to how they are laid out [here](https://primer.style/design/guides) with a heading.

The index.mdx content on /native should also be similar to /guides - just links to the content in that section (if for now it's just desktop, that's fine). It's just a catch-all so we don't dead end folks who happen to hit the url. I can help with writing some super short descriptions under the text.


> Do we need ToC as we have the sections on the right hand side?

I would remove this list of links that are on the top of this page, so that we _only_ have the ToC here
Screen Shot 2023-01-25 at 12 42 01 PM

broccolinisoup:
Asked the question about the linting rule to #accessibility ([slack ref](https://github.slack.com/archives/C0FSWLQ0Y/p1674785946370619))

emilybrick:
> I was thinking we can change the titles to have different text to make them accessible and in the meantime we can see if we can look into the doctocat theme issue and see if we can come up with a solution for it. 

☝️ sorry for the delay here, this sounds like the best option to me!

PR #370, opened 2023-01-27
Fix index page links on /design

emilybrick:
The primer.style/design index page should reflect the correct top level sections:

- Guides
- Foundations
- UI patterns
- Components

*and (eventually) Native platforms

To do
- [x] Update the index page content to include those links
- [x] Fix top level links on /guides and /accessibility

### DIFF
--- a/content/guides/index.mdx
+++ b/content/guides/index.mdx
@@ -6,43 +6,43 @@ import {Box, Link, Text} from '@primer/react'
 
 <Box display="grid" gridTemplateColumns={['1fr', null, null, null]} gridGap={4}>
   <div>
-    <Link href="/guides/introduction" sx={{fontWeight: 'bold'}}>
+    <Link href="/design/guides/introduction" sx={{fontWeight: 'bold'}}>
       Introduction
     </Link>
     <Text as="p">Get started with the principles of Primer, standards, and recommendations for designing Github.</Text>
   </div>
   <div>
-    <Link href="/guides/content" sx={{fontWeight: 'bold'}}>
+    <Link href="/design/guides/content" sx={{fontWeight: 'bold'}}>
       Content
     </Link>
     <Text as="p">How to create interfaces that follow GitHub's content guidelines.</Text>
   </div>
     <div>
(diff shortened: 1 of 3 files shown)


langermank (approved):

content/index.mdx
```diff
@@ -12,19 +12,19 @@ export default HeroLayout
     interfaces.
   </Text>
   <div>
+  <div>
```
emilybrick, line 15: Adding a link to the new top level Guides section here and removing accessibility (below), as it's now inside the Guides section

content/guides/index.mdx
```diff
@@ -6,43 +6,43 @@ import {Box, Link, Text} from '@primer/react'
 
 <Box display="grid" gridTemplateColumns={['1fr', null, null, null]} gridGap={4}>
   <div>
-    <Link href="/guides/introduction" sx={{fontWeight: 'bold'}}>
+    <Link href="/design/guides/introduction" sx={{fontWeight: 'bold'}}>
```
rezrah, line 9: Is it possible to use the Gatsby `Link` component instead of the current PRC one? The Gatsby one should include the path prefix for you in production.

🔗 [Gatsby docs for Link](https://www.gatsbyjs.com/docs/how-to/previews-deploys-hosting/path-prefix/#in-app-linking)

emilybrick: I can do this but wouldn't it be better to change them everywhere? Vs in this fix PR

rezrah (approved):
Looks good to me. 

I left a comment about using the Gatsby built-in Link component instead of the current PRC one. It would be _great_ to swap that in if possible, as it appends the `/design/` path prefix for you automatically and we can keep the href environment-agnostic. 

I'm not sure how easy it would be to swap, so I think this is fine to go as-is for now and we can do another sweep later.

content/guides/index.mdx
josepmartins: Made a PR to use Gatsby built-in links @rezrah @emilybrick 

https://github.com/primer/design/pull/373

PR #373, opened 2023-01-30
Use markdown on index.js

josepmartins:
Remove the `Text` and `Link` components from index files in favor of plain markdown language. This simplifies the code and outputs the same documentation styles for links, headings, and body text.

Related comment: https://github.com/primer/design/pull/370#discussion_r1089230569

### DIFF
--- a/content/components/anchored-overlay.mdx
+++ b/content/components/anchored-overlay.mdx
@@ -5,7 +5,7 @@ figma: https://www.figma.com/file/GCvY3Qv8czRgZgvl1dG6lp/Primer-Web?node-id=1552
 description: Anchored overlay opens an overlay relative to the anchor position.
 ---
 
-import {Box, Button, Heading, Link} from '@primer/react'
+import {Box, Button, Heading} from '@primer/react'
 
 ## Overview
 
--- a/content/components/checkbox.mdx
+++ b/content/components/checkbox.mdx
@@ -30,14 +30,13 @@ Checkboxes are capable of showing two forms of selection: checked (left) or inde
 />
 <Caption>The indeterminate state is colored in some browsers (e.g. Safari) and grey in others (e.g. Chrome).</Caption>
 
-
 ### Best practices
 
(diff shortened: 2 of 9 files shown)


josepmartins:
> Looks like we lose the link styles with Gatsby's Link component:
> 
> CleanShot 2023-01-30 at 09 53 42@2x
> 
> To preserve link styles, we could do something like this:
> 
> ```
> import {Link} from '@primer/react'
> import {Link as GatsbyLink} from 'gatsby'
> 
> <Link as={GatsbyLink>...
> ```

@colebemis That's a good point. To avoid all of this, what if we use plain markdown instead of all the React components on the index pages? That way we simplify the code, we define heading titles and get the same link styles as we do for the rest of the pages.

Example on guides index: https://github.com/primer/design/pull/373/files#diff-7b1f190bafa6d72084c51bfb62652b49d58b45afe0fe09d6a9fc3cbc4aeab982R5

Preview: https://primer-8f3317cb39-26441320.drafts.github.io/guides


Let me know if that's ok and I'll change the rest of pages 👍

rezrah:
> @colebemis That's a good point. To avoid all of this, what if we use plain markdown instead of all the React components on the index pages? That way we simplify the code, we define heading titles and get the same link styles as we do for the rest of the pages.

+1 to native markdown link _if_ that can also insert the base url prefix in prod. If not, @colebemis solution with `as` works too 👍

josepmartins:
> > @colebemis That's a good point. To avoid all of this, what if we use plain markdown instead of all the React components on the index pages? That way we simplify the code, we define heading titles and get the same link styles as we do for the rest of the pages.
> 
> +1 to native markdown link _if_ that can also insert the base url prefix in prod. If not, @colebemis solution with `as` works too 👍

native markdown links also insert the base URL prefix in prod, I fixed a broken link in this same PR that proves that. In prod this generates a `design/design/` link. 

Commit: https://github.com/primer/design/pull/373/commits/58757e416897464066de07e5f1795f2e6be31b91


Screenshot 2023-01-31 at 12 34 51

colebemis (approved):
Looks great! Nice solution @josepmartins 💖

PR #374, opened 2023-01-30
Josepmartins/test storybook component

josepmartins:
Adding a Storybook component to showcase the CSS utilities.

This is taking previous @langermank draft work on embedStoryBook components here: https://github.com/primer/design/pull/294/files#diff-f03b80256fc7ea9ed3bd0ca95452d2071e32c8ab955002a75fbd5ffcd29697a4


tobiasahlin:
What's the right context to talk about the specifics of these animations? I'd love to polish the curves of these, and align them with the other motion work that we've been doing 👀

langermank:
@tobiasahlin I think starting a Discussion around animations would be a good place to start!

simurai (approved):

PR #376, opened 2023-01-30
Copy content from primer.style/cli → primer.style/design

maximedegreve:
As part of: https://github.com/github/primer/issues/1607
We're moving all the CLI docs over and introducing a new Native section in the menu.

- Added bash to code scripts
- Unified in 3 main pages
- Changed links to be not absolute

Screenshot 2023-01-30 at 16 54 37

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,10 @@ This is currently a work in progress! Follow along on our [project board](https:
    git clone https://github.com/primer/design
    ```
 
-1. In the terminal, navigate (`cd`) to the repo directory
-1. `nvm use --default` to use the correct node version.
-1. `yarn` to install dependencies
-1. `yarn start` to start the dev server
+2. In the terminal, navigate (`cd`) to the repo directory
+3. `nvm use --default` to use the correct node version.
+4. `yarn` to install dependencies
+5. `yarn start` to start the dev server
 
 ## Deployment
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -14,10 +14,10 @@ This is currently a work in progress! Follow along on our [project board](https:
...
    ```
 
-1. In the terminal, navigate (`cd`) to the repo directory
-1. `nvm use --default` to use the correct node version.
-1. `yarn` to install dependencies
-1. `yarn start` to start the dev server
+2. In the terminal, navigate (`cd`) to the repo directory
+3. `nvm use --default` to use the correct node version.
+4. `yarn` to install dependencies
+5. `yarn start` to start the dev server
```
rezrah, line 20: Just an FYI, that leaving it as 1 will render correctly in order in GFM. Magical but confusing :D

rezrah (approved):
Looks good..

Only comment I have is that CLI is really buried in the sidebar now. It's also closed by default, rather than immediately visible through its own category. IMO it should be treated as top-level content, replacing 'native', which is somewhat superfluous here. 

Going to assume this has already been considered and is intentional. Thought I'd flag in case it wasn't. cc. @emilybrick @colebemis 

CLI at the bottom of the sidebar closed by default

PR #378, opened 2023-01-31
format: Run updated prettier on all files after version upgrade

broccolinisoup:
Since [the prettier version is changed](https://github.com/primer/design/commit/f78a870884b55e3509307b68f24d4f3151734e8f#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R14) - I realised there are quite a few formating fixes on my PR and that would be the case for future PRs. I think running prettier on all files helps pushing cleaner PRs. So this PR does that. Let me know what you think. @colebemis 

Also [gatsby-theme-doctocat version update](https://github.com/primer/design/commit/2962a1ff983512214942d66ff73cef1203d3c7aa) doesn't seem to have an updated yarn.lock so I added this one here as well.


broccolinisoup:
I'll close this ~ too many conflicts. We might open another PR if we want to run prettier on the remaning files and seems like we have [[gatsby-theme-doctocat version update](https://github.com/primer/design/commit/2962a1ff983512214942d66ff73cef1203d3c7aa)](https://github.com/primer/design/commit/09a5f4a6e1c2bf181673b4d14cc195cd2d718b55#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de) on yarn.lock so it should be good.

PR #379, opened 2023-01-31
Copy content from primer.style/mobile → primer.style/design #1614

maximedegreve:
As part of: https://github.com/github/primer/issues/1614


josepmartins (approved):
@maximedegreve you are on 🔥

PR #382, opened 2023-02-03
Fix theme menu

simurai:
This fixes the theme menu currently used for the CSS utilities. E.g. https://primer.style/design/foundations/css-utilities/colors#text

Screen Shot 2023-02-03 at 10 42 31

It might also fix the console error that @josepmartins saw:

Screenshot 2023-02-02 at 16 49 22

PR #383, opened 2023-02-03
Fix ActionMenu blank state

josepmartins:
- Fix the blank state issue when using the ActionMenu in the CSS utilities examples.
- Fix (temporary) the theme color selector to make it work with the CSS stories

Related: https://github.com/primer/design/pull/382


simurai (approved):
✨

PR #392, opened 2023-02-10
Clarify chevron direction state meaning

dylanatsmith:
I was having a brain fart about chevron directions for toggles and came looking for the obvious answer. It’s not made plain here, only inferred in screenshots.


lukasoppermann (approved):
👏

PR #408, opened 2023-02-24
Page header docs

emilybrick:
Closes https://github.com/github/primer/issues/1859

Add Page header usage guidelines to docs

### DIFF
--- a/content/components/page-header.mdx
+++ b/content/components/page-header.mdx
@@ -10,6 +10,9 @@ export default ComponentLayout
 
 ## Usage
 
+![An image of a Page header, including the title, actions, description, and navigation.](https://user-images.githubusercontent.com/586552/221928443-68004162-fe02-42d2-9ce2-7691be34c75d.png)
+
+
 Page header is used to determine the arrangement of top-level headings and to determine how elements adopt to different viewports. Page header organizes different elements that apply context and provide navigation for the current UI. 
 
 ## Anatomy 
@@ -30,7 +33,7 @@ The default way to let users navigate up in the hierarchy on Narrow viewports.
 
 Actions that appear in the Context bar specifically.
 
-### Context bar
+### Context bar area
 
 Generic slot for any component above the title region (see below). Use it for custom breadcrumbs and other navigation elements instead of Parent link
 
--- a/content/components/page-header.mdx
+++ b/content/components/page-header.mdx
@@ -37,7 +37,7 @@ By default, the [Context bar](#context-bar) is visible on narrow viewports only,
 
 ## Context bar
 
-The Context bar is intended to set context for the user on smaller viewports and provide any actions that fit within the Context bar. As a reminder, by default, the Context bar (including its actions) is only visible on narrow viewports. 
+The Context bar is intended to set context for the user on smaller viewports and provide any actions that fit within the Context bar. By default, the Context bar (including its actions) is only visible on narrow viewports. 
 
 <Box
   as="figure"
--- a/content/components/page-header.mdx
+++ b/content/components/page-header.mdx
@@ -10,32 +10,99 @@ export default ComponentLayout
 
 ## Usage
 
-![An image of a Page header, including the title, actions, description, and navigation.](https://user-images.githubusercontent.com/586552/221928443-68004162-fe02-42d2-9ce2-7691be34c75d.png)
-
-
 Page header is used to determine the arrangement of top-level headings and to determine how elements adopt to different viewports. Page header organizes different elements that apply context and provide navigation for the current UI. 
 
 ## Anatomy 
 
-Page header can include a context bar, title bar, description, and contextual navigation (pertaining to the current page/UI). 
+<img
+    src="https://user-images.githubusercontent.com/586552/221939296-410d3b17-ccbb-48ae-b1cf-c3a8cdcfcaa5.png"
+    role="presentation"
+    width="993"
+    alt="Page header anatomy"
+/>
+
+Page header can include a context bar, title bar, description, and contextual navigation (pertaining to the current page). See below for more details.
 
 ## Context bar
 
 By default, the Context bar (including its actions) is only visible on Narrow viewports. This is intended to set context for the user on smaller viewports and provide any actions that fit within the context bar.
 
-### Parent link
+<Box
+  as="figure"
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column-reverse', 'column-reverse', 'column-reverse', 'column-reverse', 'row']}
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+  <Box as="figcaption" mt="1">
+    <Text as="h4" fontSize={2} fontWeight="bold" m={0}>
+      Parent link
+    </Text>
+    <Text as="p" mt="0">
+        The default way to let users navigate up in the hierarchy on Narrow viewports.
+    </Text>
+    <Text as="p" mt="0">
+        Responsive strategy: On Narrow viewports, the Parent link touch target expands beyond its element boundaries.
+    </Text>
+  </Box>
+  <img
+    src="https://user-images.githubusercontent.com/586552/221946033-5af00289-af9e-4aae-9b0b-af0ecf3a12b9.png"
+    role="presentation"
+    width="456"
+  />
+</Box>
 
-The default way to let users navigate up in the hierarchy on Narrow viewports.
+<Box
+  as="figure"
+  display="flex"
+  alignItems="flex-start"
+  justifyItems="space-between"
+  flexDirection={['column-reverse', 'column-reverse', 'column-reverse', 'column-reverse', 'row']}
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+  <Box as="figcaption" mt="1">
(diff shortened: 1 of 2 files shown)


content/components/page-header.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+The default way to let users navigate up in the hierarchy on Narrow viewports.
+
+**Responsive strategy**: On Narrow viewports, the ParentLink touch target expands beyond its element boundaries, to match with the minTouchTarget design token.
+
+### Context bar actions
+
+Actions that appear in the Context bar specifically.
+
+### Context bar
```
josepmartins, line 33: The markdown lints fail as there are two identical `Context bar` titles

emilybrick: ah thanks for catching!

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
+---
+title: Page header
+reactID: drafts_page_header
```
emilybrick, line 3: this tab still isn't showing up

emilybrick: fixed ❤️ thank you @colebemis

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
...
+<img 
+    src="https://user-images.githubusercontent.com/586552/222216317-c8478ab0-6541-49c6-9140-bc884e9a8cd4.png"
+    role="presentation"
+    width="877"
+    alt="Page header context bar in a narrow viewport"
+/>
+
+## Title bar
+
+The Title bar can include a static page title, actions, and a parent link that by default is only visible on narrow viewports.
```
mperrotti, line 131: It would be nice to explain what kinds of actions go in the title bar.

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
...
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+  <Box as="figcaption" mt="1">
+    <Text as="h3" fontSize={2} m={0}>
+      Default title variant
+    </Text>
+    <Text as="p" mt="0">
+        The default way to let users navigate up in the hierarchy on narrow viewports.
```
mperrotti, line 148: So this title is also a link? It's not really clear in the example image. It just looks like this is the "default" text size/style for the title.

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
...
+    </Text>
+  </Box>
+  <img
+    src="https://user-images.githubusercontent.com/586552/222215097-79bb0d77-af5c-4199-81a1-65d98a7a0556.png"
+    role="presentation"
+    width="456"
+  />
+</Box>
+
+### Title bar responsive strategy
```
mperrotti, line 208: I think we can remove this section too.

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
...
+
+On narrow viewports, all title variants appear as medium.
+
+
+## Title bar visual regions
```
mperrotti, line 213: I think we can combine this content with the title variants

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
...
+    </Text>
+  </Box>
+  <img
+    src="https://user-images.githubusercontent.com/586552/222194459-daf8c4b8-0452-40a2-bdf3-2b74be06e7ef.png"
+    role="presentation"
+    width="456"
+  />
+</Box>
+
+### Title bar visual regions responsive strategy
```
mperrotti, line 265: I think we can remove this section too.

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
...
+
+On narrow viewports, all visual regions remain visible by default.
+
+## Title bar action regions
```
mperrotti, line 269: I think we can combine this content with the title variants and leading/trailing visuals.

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
+---
+title: Page header
+reactId: drafts_page_header
+description: Page header determines the top-level headings of a UI
```
mperrotti, line 4: It applies to way more than just headings. Maybe we can try something like

```suggestion
description: Page header arranges top-level contextual information and controls for the page
```

content/components/page-header.mdx
emilybrick: I'm not sure what you mean here ^

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
...
+figma: https://www.figma.com/file/Ee0OrXuOLXMDqUW83EnDHP/PageHeader-(FY23-Q1)?node-id=0%3A1&t=8cKRbkNYhwbqEoHj-1
+---
+
+import {Box, Text} from '@primer/react'
+import ComponentLayout from '~/src/layouts/component-layout'
+export default ComponentLayout
+
+## Usage
+
+Page header is used to determine the arrangement of top-level headings and to determine how elements adopt to different viewports. Page header organizes different elements that apply context and provide navigation for the current UI.
```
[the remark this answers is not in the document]
emilybrick: Okay - just want to note here we should also update the component metadata, where this comes from ^

content/components/page-header.mdx
```diff
@@ -0,0 +1,330 @@
...
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+  <Box as="figcaption" mt="1">
+    <Text as="h3" fontSize={2} m={0}>
+      Context bar actions
+    </Text>
+    <Text as="p" mt="1">
+        Actions that appear in the Context bar specifically.
```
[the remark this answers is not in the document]
emilybrick: I think the only reason this distinction (taken from Figma) is because they are actions that live specifically in the context bar, vs the actions in the title bar (which appear below). I don't believe there are any differences between the actions here vs the title bar - except that they pertain to their context

content/components/page-header.mdx
```diff
@@ -0,0 +1,295 @@
...
+    role="presentation"
+    width="960"
+    alt="Page header component"
+/>
+
+## Anatomy 
+
+Page header can include a context bar, title bar, description, and contextual navigation (pertaining to the current page). 
+
+By default, the [context bar](#context-bar) is visible on narrow viewports only, the [title bar](#title-bar) is visible on all viewports, and the back button and trailing action are visible on regular viewports.
```
broccolinisoup, line 29: The back button is renamed to `LeadingAction` so it should be "...and the leading and trailing action are visible on regular viewports."

Would it be also possible to update this on Figma and take a new screenshot please? 🙇🏻‍♀️

content/components/page-header.mdx
```diff
@@ -0,0 +1,295 @@
...
+    src="https://user-images.githubusercontent.com/586552/222194469-85131849-4a7a-41e6-abca-040d042110df.png"
+    role="presentation"
+    width="456"
+  />
+</Box>
+
+## Title bar
+
+The title bar can include a static page title, actions, and a parent link that by default is only visible on narrow viewports.
+
```
broccolinisoup, line 121: This might be a copy-paste mistake?

content/components/page-header.mdx
```diff
@@ -0,0 +1,295 @@
...
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+  <Box as="figcaption" mt="1">
+    <Text as="h3" fontSize={2} m={0}>
+      Back button
+    </Text>
+    <Text as="p" mt="0">
+        A back button can be used as a leading action for local navigation. On narrow viewports, use parent link instead.
```
broccolinisoup, line 262: Back button -> leading visual :)

content/components/page-header.mdx
```diff
@@ -0,0 +1,295 @@
...
+  m={0}
+  mb={3}
+  sx={{gap: 3}}
+>
+  <Box as="figcaption" mt="1">
+    <Text as="h3" fontSize={2} m={0}>
+      Trailing visual
+    </Text>
+    <Text as="p" mt="0">
+        Trailing visual and trailing text can display auxiliary information. They're placed at the right of the item, and can denote status, privacy details, etc.
```
broccolinisoup, line 237: Should we also add that the leading and trailing visuals remain visible on all viewports?

content/components/page-header.mdx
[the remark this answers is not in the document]
mperrotti: If the context bar is only visible on narrow viewports, what do we do with those actions for wider viewports? That might be good info for here.

Neither of us was very involved in the design of this component, so we might have to just merge and leave things vague until we get some answers.

content/components/page-header.mdx
mperrotti: If we can't figure out which actions go in the title bar vs the context bar, we can just merge and wait until we figure out an answer about that.

content/components/page-header.mdx
emilybrick: Fixed! apologies - you told me that and i thought it was the other way around. Should be fixed to 'leading action' now

content/components/page-header.mdx
emilybrick: fixed now

content/components/page-header.mdx
emilybrick: done ✔️

PR #419, opened 2023-03-07
Simplify component status page

colebemis:
## Before

CleanShot 2023-03-07 at 15 23 46@2x

## After

CleanShot 2023-03-07 at 15 20 43@2x


emilybrick (approved):
Such a nice change - thank you! I don't think the description were giving us much. Once we have all the components in /design I wonder if we should link to them from the first column.

PR #428, opened 2023-03-10
Add interface guidelines for TextInput

mperrotti:


content/components/text-input.mdx
```diff
@@ -0,0 +1,156 @@
...
+
+<img
+  width="960"
+  role="presentation"
+  src="https://user-images.githubusercontent.com/2313998/224445201-7f81cf2c-992c-470b-b702-4b45d57c3234.png"
+/>
+
+A trailing action may be used to render a single action that affects the input. A trailing action is most commonly used for a button that clears the input.
+
+Ideally the trailing action button is an icon without a text label. It is also possible to render the button with a text label, but it should be avoided whenever possible.
```
emilybrick, line 117: is there a time when we would actually encourage this? trying to think of an example

content/components/text-input.mdx
```diff
@@ -0,0 +1,156 @@
...
+
+### Monospace
+
+<img
+  width="960"
+  role="presentation"
+  src="https://user-images.githubusercontent.com/2313998/224445199-c8f27a2b-18e3-4ac7-9764-1d32b1fdc9c7.png"
+/>
+
+The value of the text input may be rendered in a monospace font. Use this variant for inputs related to code. For example, setting a personal access token in a configuration.
```
emilybrick, line 105:
```suggestion
The value of the text input may be rendered in a monospace font. Use this variant for inputs related to code, such as setting a personal access token in a configuration.
```

content/components/text-input.mdx
```diff
@@ -0,0 +1,156 @@
+---
+title: Text input
+description: A text input is used to set a value that is a single line of text.
```
emilybrick, line 3:
```suggestion
description: Text input is used to set a value that is a single line of text.
```

emilybrick (approved):
These look great. Thank you!

content/components/text-input.mdx
mperrotti: I couldn't think of an example either, but we support it in the implementation.

PR #436, opened 2023-03-16
Link documentation

emilybrick:
These guidelines follow: https://github.com/primer/design/blob/main/content/components/component-documentation-guidelines/component-MVP-documentation-guidelines.md

They can absolutely be built upon, I see this more as a starting point for us and a way to surface links in docs (pull in /react + figma docs as well). 

This also pulls in some of the link accessibility documentation that @lukasoppermann wrote

### DIFF
--- a/content/components/link.mdx
+++ b/content/components/link.mdx
@@ -46,7 +46,7 @@ The link component can be displayed in default, muted, or blue colors. Links can
 
 Links play a key part in your experience on the web, but without proper consideration they can be frustrating to use, skipped over, or completely unnoticed.
 
-For screen reader assistive technology, links and buttons are expected to function differently from each other. If a link is activated and does not do what was expected, that can be disorienting and frustrating.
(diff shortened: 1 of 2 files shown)


content/components/link.mdx
```diff
@@ -0,0 +1,265 @@
...
+    role="presentation"
+    src="https://user-images.githubusercontent.com/586552/225489086-e604fbc8-e449-416e-bd26-cc80408c5d8d.png"
+   />
+</Box>
+
+## Accessibility
+
+Links play a key part in your experience on the web, but without proper consideration they can be frustrating to use, skipped over, or completely unnoticed.
+
+For screen reader assistive technology, links and buttons are expected to function differently from each other. If a link is activated and does not do what was expected, that can be disorienting and frustrating.
```
mperrotti, line 49:
```suggestion
For assistive technologies like screen readers, links and buttons are expected to function differently from each other. If a link is activated and does not do what was expected, that can be disorienting and frustrating.
```

I think this may affect more than just screen readers, so I rephrased this to be slightly more inclusive of other ATs.

mperrotti:
I was finally able to look at the preview deployment. Merge it!

Issue #432, opened 2023-03-14
Automate Components index page


emilybrick:
**Problem:** 
We have an [index page for our components](https://primer.style/design/components), but right now it must be manually updated and is difficult to keep in sync

<img width="1042" alt="Screenshot 2023-03-14 at 2 33 44 PM" src="https://user-images.githubusercontent.com/586552/225104017-74e5e2e8-0e78-4302-93ce-9c03d340218b.png">

**Goal:** 
Automate the list of components + descriptions that show up on [primer.style/design/components](https://primer.style/design/components)

**Why this matters:** 
Though you can't click on `Components` from the navigation, there will likely be many places where we will want to link to this 'landing' page from, such as the main Primer homepage. 

**Exit criteria:**
- List of components that appear on https://primer.style/design/components matches the list of components in our documentation that appear in the navigation. 
- Each component should have name and description (thumbnails are nice to have, but not necessary for v1)

PR #438, opened 2023-03-17
Generate content for index pages

colebemis:
Generates content for index pages so we don't have to keep them up-to-date manually

#### Markdown

```
---
title: Components
---

<ChildPages of="Components" />
```

#### Rendered page

CleanShot 2023-03-17 at 15 01 22@2x

Closes https://github.com/primer/design/issues/432


src/components/child-pages.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <Box sx={{display: 'grid', gap: 5}}>
+      {childItems.map(item => (
+        <Box key={item.url} sx={{display: 'grid', gap: 1}}>
+          <Heading as="h3" sx={{fontSize: 2}}>
+            <Link as={GatsbyLink} to={item.url}>
+              {item.title}
+            </Link>
+          </Heading>
+          <Text as="p" sx={{m: 0, maxWidth: '60ch'}}>
```
emilybrick, line 57: can we increase the maxWidth here? Feels a little narrow on wide screens
before:
Screenshot 2023-03-21 at 7 44 09 PM


after (80):
Screenshot 2023-03-21 at 7 43 58 PM


Separately - we should center this content but this is outside of this PR

emilybrick (approved):

PR #439, opened 2023-03-17
Fix table overflow on component guideline pages

colebemis:
### Before

CleanShot 2023-03-17 at 15 13 10@2x

### After

CleanShot 2023-03-17 at 15 12 39@2x

Closes https://github.com/primer/react/issues/3013

### DIFF
--- a/src/layouts/component-layout.tsx
+++ b/src/layouts/component-layout.tsx
@@ -53,7 +53,7 @@ export default function ComponentLayout({pageContext, children, path}) {
               </>
             ) : null}
           </Box>
-          <Box>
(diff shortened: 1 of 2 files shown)


src/layouts/component-layout.tsx
```diff
@@ -53,7 +53,7 @@ export default function ComponentLayout({pageContext, children, path}) {
               </>
             ) : null}
           </Box>
-          <Box>
+          <Box sx={{minWidth: 0}}>
```
jonrohan, line 56: `min-width: 0` ftw 🙌🏻

jonrohan (approved):

PR #440, opened 2023-03-17
Add SSR compatibility to component lifecycle criteria

colebemis:


lesliecdubs (approved):
I'm good with this change assuming we do fast-follow ups to address `TreeView` and `ActionList` which are already in Beta but don't meet the new SSR requirement.

More discussion for reference: https://github.slack.com/archives/C02NUUQ9C30/p1679088466733799

PR #451, opened 2023-03-23
Update numeric table cell guidance

mperrotti:
Mentions using tabular figures for easier scanning/comparing in a column.


joshblack (approved):
🎉

PR #459, opened 2023-03-31
Add interface guidelines for RelativeTime

mperrotti:


content/components/relative-time.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+For more information about Primer's patterns for date and tme formatting, see the "[Dates and times](/ui-patterns/dates-and-times)" UI pattern guidelines.
+
+## Options
+
+### Verbosity
+
+**Auto**: "Apr 1, 2014"
+
+**Micro**: "9y"
```
keithamus, line 20: The micro pattern is not recommended due to issues with screen readers mispronouncing "9m" as "9 meters". We should discourage its use.

keithamus:
Is it worth just setting up a redirect from `RelativeTime` to `Dates & Times`? It covers all of our guidance related to this.

content/components/relative-time.mdx
josepmartins: Agree with this, we cover that in the Date & Times docs: https://primer.style/design/ui-patterns/dates-and-times#responsive-design-and-micro-format

mperrotti: @josepmartins - do you want to create an issue to remove the `"micro"` value as an option for the `format` prop?

josepmartins (approved):
LGTM! 🎉

keithamus (approved):

PR #462, opened 2023-04-03
Add Rails components to the new docsite

camertron:
This PR adds Rails components to the new docsite! 🎉 It's dependent on [this PVC PR](https://github.com/primer/view_components/pull/1908) for the data necessary to render all the pages. A couple of things to consider when reviewing:

1. The data in PVC contains mustache tags for linking to components, system arguments, etc. This is because it's very hard for PVC to reliably guess the correct URL to a component exposed by the new Gatsby site. There is code in this PR that can map a `railsId` (i.e. the component's fully-qualified Ruby constant) to a page ID, which makes linking possible.
2. The PVC docs for system arguments are written in inline markdown above the definition of the `BaseComponent` class. While I could have simply copy/pasted them into the new docsite, I think it makes sense for the source of truth for these docs to remain PVC. Accordingly, system arguments docs in markdown format are copied verbatim into info_arch.json and rendered using `<ReactMarkdown>` in the new system arguments layout.
3. Any component that has a Rails implementation but _not_ a React one now has a very basic overview section with a single warning banner:

    image
4. Lookbook previews are rendered by pointing an iframe at primer.style/view-components in prod and localhost:4000 in dev. Currently the new docsite does not point to Lookbook preview environments.
5. Both `remark` and `remark-gfm` have been added as dependencies so the code examples in the system arguments docs will render correctly.
6. In the data generated by PVC, sub-components are listed under the `subcomponents` key of their parent component. The graphql query in the Rails layout fetches two nesting levels of these, which should be enough since I don't believe any sub-components themselves contain sub-components. However if we ever need a third nesting level, I would recommend a flatter info hierarchy because graphql does not particularly like or support recursive data structures.


src/layouts/rails-component-layout.tsx
```diff
@@ -1,33 +1,253 @@
...
+  const slugger = new GithubSlugger()
+
+  for (const subcomponent of subcomponents) {
+    tableOfContents.items.push({
+      url: `#${slugger.slug((subcomponent as any).name)}`,
+      title: (subcomponent as any).name
+    })
+  }
+
+  const renderSubComponents = (subcomponents) => {
```
colebemis, line 237: In general, whenever you find yourself writing a function called `render*` in React, it's a good indication that it should be a component. I'd pull this out into a `RailsSubComponents` component

colebemis (commented):
The code changes look good. I just left a few minor comments. Unpkg is down so I can't verify the changes in the browser yet.

src/@primer/gatsby-theme-doctocat/nav.yml
```diff
@@ -106,6 +106,8 @@
           url: /foundations/icons/octovisuals
         - title: Design guidelines
           url: /foundations/icons/design-guidelines
+    - title: System arguments
```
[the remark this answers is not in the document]
camertron: I would really like to keep everything in one place if we can.

colebemis:
[Minor] I noticed that a lot of Rails descriptions are redundant:

CleanShot 2023-04-06 at 13 41 28@2x

src/layouts/rails-component-layout.tsx
```diff
@@ -41,37 +268,131 @@ export default function RailsComponentLayout({data}) {
...
               sx={{display: 'inline-flex', gap: 1, alignItems: 'center'}}
-              href={data.sitePage.context.frontmatter.railsUrl}
+              href={railsUrl}
               target="_blank"
               rel="noopener noreferrer"
             >
-              {componentName}
+              {name}
               <LinkExternalIcon />
             </Link>
```
colebemis, line 319: This link was a place holder. I think we can remove it now.

src/layouts/rails-component-layout.tsx
```diff
@@ -1,33 +1,260 @@
...
+}
+
+function RailsComponent({data, showPreviews, parentRailsId}) {
+  const {props, slots, methods, previews} = data
+
+  return(<>
+    <RailsComponentArguments props={props} parentRailsId={parentRailsId}/>
+    <RailsComponentSlots slots={slots} parentRailsId={parentRailsId}/>
+    <RailsComponentMethods methods={methods} parentRailsId={parentRailsId}/>
+    <RailsComponentPreviews previews={previews} showPreviews={showPreviews}/>
```
colebemis, line 194: Is there a reason you put the examples under the API reference? In the React pages, we have examples before props.

colebemis (approved):
I left a few minor comments, but overall this looks great! Amazing work, @camertron 💖

src/layouts/rails-component-layout.tsx
camertron: I think I was following the order we have things in our legacy Gatsby docsite, but I'd be more than happy to swap things around 👍

camertron:
@colebemis ah yeah good catch re: the redundant descriptions. I wasn't sure what to do about them because in some instances they provide additional information that the description in the frontmatter doesn't. In some cases however they're identical. I could potentially skip rendering the Rails one if it's identical to the frontmatter one?

colebemis:
>  I could potentially skip rendering the Rails one if it's identical to the frontmatter one?

I think it's okay to ship this initial version with the descriptions. Eventually, we'll probably want to delete the redundant descriptions in the Rails source code.

src/@primer/gatsby-theme-doctocat/nav.yml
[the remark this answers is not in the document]
emilybrick: Hey - apologies was out last week. I just noticed this merged 😬. I think it's a bit confusing to have the Rails development specific documentation in here, rather than in the repo. It's unclear from the nav (and title) that this is geared specifically towards Rails, whereas our other documentation that is framework-specific and living in the docs is clearly labeled as such (for ex - [React dev guidelines](https://primer.style/design/guides/development/react), [CSS utilities](https://primer.style/design/foundations/css-utilities/getting-started), and the React/Rails specific sections of components.

PR #468, opened 2023-04-13
Updates TabNav docs and adds TabPanel docs

mperrotti:
The previous TabNav docs contained guidance that was intended for TabPanels.

### DIFF
--- a/content/components/tab-nav.mdx
+++ b/content/components/tab-nav.mdx
@@ -1,14 +1,15 @@
 ---
 title: Tab nav
-description: Tab nav contains a set of links that let users switch between views in the same context.
+description: Tab nav contains a set of links that let users navigate between different views in the same context.
 reactId: tab_nav
-rails: https://primer.style/view-components/components/alpha/tabnav
+railsId: Primer::Alpha::TabNav
 componentId: tab_nav
 figma: https://www.figma.com/file/GCvY3Qv8czRgZgvl1dG6lp/Primer-Web?node-id=20366-72067&t=kKGC85STU5bcL0zg-4
 ---
 
 import ComponentLayout from '~/src/layouts/component-layout'
 export default ComponentLayout
+import {Box} from '@primer/react'
 
 <img
   width="960"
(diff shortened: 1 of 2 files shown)


mperrotti:
If @ericwbailey confirms that arrow keys should not move focus or change the active tab in a TabNav, we should create an issue to fix the [@primer/react implementation](https://primer.style/react/storybook/?path=/story/components-tabnav--default).

content/components/tab-panels.mdx
```diff
@@ -1,14 +1,91 @@
...
+## Accessibility
+
+### Keyboard navigation
+
+<Box display="flex" flexDirection={['column', 'column', 'column', 'column', 'row']} sx={{gap: 3}}>
+  <Box flex={1}>
+
+Tabs, like all interactive elements, can be activated using a cursor or keyboard.
+
+When a user moves focus into the `tablist` using the Tab key, the active tab should be focused. A `tablist` is only a single Tab stop.
```
ericwbailey, line 67:
```suggestion
When a user moves focus into the `tablist` using the Tab key, the active tab should be focused. A `tablist` is only a single Tab stop. Once focused, navigation through the `tablist` can then conducted with Left, Right, Home, and End keypresses. 
```

Expanded on this a bit to communicate what is expected after the `tablist` receives focus.

content/components/tab-nav.mdx
```diff
@@ -18,26 +19,20 @@ export default ComponentLayout
 
 ## Usage
 
-A tab nav is used to organize related content into different views within the same context.
+A tab nav lets users navigate between related views in the same context. Each "tab" links to a different view that is shown when a user selects a tab.
+
+A tab nav is always a set of links, and each tab should have a unique URL.
```
ericwbailey, line 24: What do you think of having a sentence here that includes a link to tab panels, and the tab panels usage section having a sentence that includes a link to tab nav?

My thinking is it could help pre-empt folks who don't know the difference between the two components' use cases and provides a link for them to self-serve.

ericwbailey (approved):
@mperrotti Confirming that arrow keys should not factor into TabNav.

Also added a suggestion and a question. Otherwise this looks good to me!

PR #469, opened 2023-04-14
Fix Figma layout missing Rails tabs 

josepmartins:
- Fix Figma layout to use RailsId instead of RailsURL
- Update `RelativeTime` component to use the new layout


danielguillan (approved):

lukasoppermann (approved):
👍

PR #470, opened 2023-04-19
Update tree-view.mdx

andrialexandrou:
In Office Hours April 19 we discussed that tree view should not be used for website navigation.

Will add
- detailed explanation of why not to use for site nav (SR experience requires going in and out of an interaction mode)
- recommendation for what to do instead of using this for site nav


ericwbailey (approved):

andrialexandrou:
I'm dragging my feet. Merging as is.

PR #478, opened 2023-04-25
Clarify why a TreeView should not be used for nav

ericwbailey:
This PR addresses [a request in Slack](https://github.slack.com/archives/C01L618AEP9/p1682453128763549?thread_ts=1682450604.837379&cid=C01L618AEP9) by @howdyray. It adds a paragraph to explain usability expectations for the TreeView component, and how there may be a mismatch in expectation if used for static navigation.

### DIFF
--- a/content/components/tree-view.mdx
+++ b/content/components/tree-view.mdx
@@ -239,6 +239,8 @@ If a tree view is truly the best pattern for your use-case and none of the sugge
 
 A tree view solves a very specific problem. It's not a multi-purpose tool like an [action list](/action-list) or meant for site navigation.
 
+Tree views are used to communicate a hierarchical list of items, and allow a user to navigate through, select, and take action on one or more items. A comparable expereince would be a list of files in a code editor, or an operating system file explorer. While they may visually look like navigation, tree views have specific interaction modes and expectations for assistive technology that differs from other patterns. Trying to use a tree view for something that looks visually apprpriate, but is not functionally inline with its intended use case may cause confusion or an unusable experience, especially if the user cannot see the screen.
+
 <Text as="p" m={0}>
   Before reaching for a tree view, first make sure that:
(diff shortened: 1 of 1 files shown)


mperrotti (approved):
Thank you!

This can merge as-is, but it would be even better if in the "Accessibility" section we could explain how using a tree view as navigation creates a worse experience for ATs.

content/components/tree-view.mdx
```diff
@@ -239,6 +239,8 @@ If a tree view is truly the best pattern for your use-case and none of the sugge
 
 A tree view solves a very specific problem. It's not a multi-purpose tool like an [action list](/action-list) or meant for site navigation.
 
+Tree views are used to communicate a hierarchical list of items, and allow a user to navigate through, select, and take action on one or more items. A comparable expereince would be a list of files in a code editor, or an operating system file explorer. While they may visually look like navigation, tree views have specific interaction modes and expectations for assistive technology that differs from other patterns. Trying to use a tree view for something that looks visually apprpriate, but is not functionally inline with its intended use case may cause confusion or an unusable experience, especially if the user cannot see the screen.
```
joshblack, line 242:
```suggestion
Tree views are used to communicate a hierarchical list of items, and allow a user to navigate through, select, and take action on one or more items. A comparable experience would be a list of files in a code editor, or an operating system file explorer. While they may visually look like navigation, tree views have specific interaction modes and expectations for assistive technology that differs from other patterns. Trying to use a tree view for something that looks visually appropriate, but is not functionally inline with its intended use case may cause confusion or an unusable experience, especially if the user cannot see the screen.
```

ericwbailey:
@mperrotti I just pushed up a note about it, if you don't mind giving that a second review.

mperrotti (approved):

PR #480, opened 2023-05-02
Update pointing device guidance to accommodate multimodal input

ericwbailey:
This PR updates the [Responsive section of our documentation](https://primer.style/design/foundations/responsive), specifically the "Pointing device" guidance. 

## The short version

- Devices don't always accurately report or respond to pointer media queries, and can use input devices that aren't always considered.
- A standard minimum `24px` interactive area for all input use cases can address this.

### Context

The incoming changes remove the support table based on pointer device fidelity. The reasoning for this is that a device's pointing capability is not necessarily a binary state, and may shift in flight in situations where multimodal input is present. 

Examples of multimodal input scenarios are devices that can simultaneously support: 

- Pointing cursors (mice and touchpads), 
- Direct touch gestures (finger or capacitive device on screen), 
- Keyboard, and 
- Voice input.

Nearly all contemporary devices support multimodal input, including desktops, laptops, tablets, and smartphones. 

Not all devices will reliably report the presence of an augmented pointing device, or successfully update the media query conditions in realtime if a new pointing device is supplied (ex: plugging a USB mouse into a tablet and clicking on things may be reported as a touch gesture).

Broadly-speaking, there are two types of this use case: 

1. User-provided augmentation that is technically supported by device hardware and software, but not actively promoted in the mainstream discourse, and
2. Manufacturer supported augmentation that is promoted as a competitive feature.

Examples of two interactions derived from this use case are:

1. Using a hardware device that mimics a mouse with an Android phone. 
2. An iPad set to Sidecar mode as a second display.

User input augmentation sometimes manifests in the form of an assistive technology need. An example of this is an individual with severe motor control conditions who is unable to perform accurate touch gestures over a sustained period of time. Here, they may use a device that translates short physical movements into emulated mouse cursor movements. 

For this use case in particular, touchscreen devices are often utilized given their lower cost, portable nature, ability to be mounted on devices such a wheelchairs, and other manufacturer-provided accessibility features. 

### Guidance

Multimodal input is a real-world use case consideration, and browsers may not be privy to the truth of a user's hardware-level input operation. Because of these considerations, we should be not relying on `fine` or `coarse` as determinants of rendered UI size. 

This approach may actually work against the use case it was intended for. The updated approach advises readers to instead design with WCAG-compliant interactive target sizes for **all** potential interaction modes.

The updated guidance also updates to recommend a minimum constant of `24px` for all possible input scenarios. This guidance aligns with the upcoming WCAG 2.2 update that is slated to make SC 2.5.5: Target Size a AA-level concern.

## Additional reading

- [Interaction Media Features and Their Potential (for Incorrect Assumptions)](https://css-tricks.com/interaction-media-features-and-their-potential-for-incorrect-assumptions/)
- [Getting touchy: Everything you (n)ever wanted to know about touch and pointer events](https://patrickhlauke.github.io/getting-touchy-presentation/#)
- [Detecting touch: it's the 'why', not the 'how'](https://hacks.mozilla.org/2013/04/detecting-touch-its-the-why-not-the-how/)

### DIFF
--- a/content/foundations/responsive.mdx
+++ b/content/foundations/responsive.mdx
@@ -50,7 +50,7 @@ Because of this, pointing device media queries such as `coarse` or `fine` are un
 When creating interactive targets in your design, ensure that they are large enough to easily tap or click on. The AA accessibility standard GitHub strives for requires a minimum target size of `24px`.
 However, it is recommended to aim for the AAA standard when possible. For more information on Target Size at the AAA level, refer to the [W3C documentation](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html).
 
-A common example of a hit target that may not meet the AAA standard on mobile is our medium size button, which is only 32px in height. To improve its touch target, the height could be increased by adjusting its padding. This will increase the touch target's overall size without altering the visual appearance of the button.
+A common example of a hit target that may not meet the AAA standard on mobile is our medium size button, which is visually only 32px in height. To improve the button's touch target size without viusually adjusting its appearance, the touch target area could be increased by adjusting the button's padding. This will increase the button's touch target's overall area without needing to adjust the button's visual size.
 
 <DoDontContainer>
   <Do>
--- a/content/foundations/responsive.mdx
+++ b/content/foundations/responsive.mdx
@@ -50,7 +50,7 @@ Because of this, pointing device media queries such as `coarse` or `fine` are un
 When creating interactive targets in your design, ensure that they are large enough to easily tap or click on. The AA accessibility standard GitHub strives for requires a minimum target size of `24px`.
 However, it is recommended to aim for the AAA standard when possible. For more information on Target Size at the AAA level, refer to the [W3C documentation](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html).
 
-A common example of a hit target that may not meet the AAA standard on mobile is our medium size button, which is only 32px in height. To improve its touch target, the height must be increased without altering the visual appearance of the button.
+A common example of a hit target that may not meet the AAA standard on mobile is our medium size button, which is only 32px in height. To improve its touch target, the height could be increased by adjusting its padding. This will increase the touch target's overall size without altering the visual appearance of the button.
 
 <DoDontContainer>
   <Do>
--- a/content/foundations/responsive.mdx
+++ b/content/foundations/responsive.mdx
@@ -41,22 +41,14 @@ Todo: Link to Figma Templates
 
 ### Pointing device
 
-Use the device characteristics through [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries) to adapt the experience accordingly for different device scenarios. The table below shows some possible device scenarios using `pointer` and `hover` media features:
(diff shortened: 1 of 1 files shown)


content/foundations/responsive.mdx
```diff
@@ -41,23 +41,15 @@ Todo: Link to Figma Templates
...
-| &nbsp;             |                `pointer: coarse`                 |   `pointer: fine`    |
-| ------------------ | :----------------------------------------------: | :------------------: |
-| **`hover: none`**  |            Smartphones, touch screens            | Stylus-based screens |
-| **`hover: hover`** | Virtual reality headsets, video-game controllers |   Mouse, touch pad   |
+Because of this, pointing device media queries such as `coarse` or `fine` are unreliable. Avoid relying on this type of query to conditionally modify parts of an interface.
 
 #### Minimum target
 
-When creating interactive targets in your design, ensure that they are large enough to easily tap or click on. The AA accessibility standard GitHub strives for requires a minimum target size of `32px` for touch `pointer: coarse` and `24px` for mouse/stylus `pointer: fine`.
+When creating interactive targets in your design, ensure that they are large enough to easily tap or click on. The AA accessibility standard GitHub strives for requires a minimum target size of `24px`.
```
langermank, line 50: This is great! So that means even our [smallest control sizing tokens](https://primer.style/primitives/storybook/?path=/story/size-functional-control--x-small) actually cover this without the additional media query I believe?

content/foundations/responsive.mdx
```diff
@@ -41,23 +41,15 @@ Todo: Link to Figma Templates
...
 However, it is recommended to aim for the AAA standard when possible. For more information on Target Size at the AAA level, refer to the [W3C documentation](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html).
 
-| Input             | Minimum Size (AA) | Recommended size (AAA) |
-| ----------------- | :---------------: | :--------------------: |
-| `pointer: coarse` |       32px        |          44px          |
-| `pointer: fine`   |       24px        |          24px          |
-
 A common example of a hit target that may not meet the AAA standard on mobile is our medium size button, which is only 32px in height. To improve its touch target, the height must be increased without altering the visual appearance of the button.
```
langermank, line 53: I noticed this was left in. Would you recommend we continue to add an "invisible" touch target using media queries even though its unreliable in many cases? Is it harmful to leave in place?

langermank (approved):
Thanks for such a detailed explanation, I learned a lot!

content/foundations/responsive.mdx
ericwbailey: I think so, yes!

content/foundations/responsive.mdx
ericwbailey: Oh my gosh, this is totally a case of me assuming something and it could definitely be clearer. I just pushed up a copy tweak to clarify, does that read better to you now?

langermank: I _think_ what you're saying is that the only way to confidently improve the touch target is to change the visual appearance of the `Button` by making it larger. So I think what you pushed up partly is more clear but also might not be accurate 😆 which is fine, I would just remove the bit about it not changing the visual appearance (unless I'm reading this wrong).

langermank (approved):
Looks good!

PR #487, opened 2023-05-12
Split up ui-patterns/forms into tabs

camertron:
Fixes https://github.com/github/primer/issues/2253

This PR accomplishes the following:

1. Adds tabs to the "UI patterns -> Forms" page. There are now tabs for interface guidelines ("Overview"), React, and Rails.
2. Adds a warning to the top of each Rails form component page urging the use of the framework instead of the standalone component.

### DIFF
--- a/content/ui-patterns/rails-forms.mdx
+++ b/content/ui-patterns/rails-forms.mdx
@@ -1,0 +1,39 @@
+---
+title: Forms in Rails
+description: Primer contains a framework for declaratively building Rails forms. It was designed specifically for accessibility and consistency with the Primer design system.
+---
+
+import {Heading, Link} from '@primer/react'
+import {LinkExternalIcon} from '@primer/octicons-react'
+
+## Introduction
+
+Creating accessible forms can be challenging. Forms are a common source of accessibility issues, and can hinder someone from successfully completing and submitting a form. This can be true even for people who do not use assistive technologies.
+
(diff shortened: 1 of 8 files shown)


content/ui-patterns/rails-forms.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: Forms in Rails
```
colebemis, line 2: Would love to get @emilybrick's input on whether this is the correct spot in the IA for this documentation

src/layouts/rails-component-layout.tsx
```diff
@@ -313,20 +315,24 @@ export default function RailsComponentLayout({data}) {
...
-              <Link href={railsUrl}>Primer ViewComponents documentation for {name}</Link>.
-            </Note>
-
             <Box sx={{display: 'flex', gap: 2, mb: 4}}>
               <Label size="large">v{data.primerRailsVersion.version}</Label>
               <StatusLabel status={sentenceCase(status)} />
               <AccessibilityLabel a11yReviewed={a11y_reviewed} short={false} />
             </Box>
 
+            {/* @ts-ignore */}
```
emilybrick, line 324: I don't think we should be using a warning here, as this is knowledge the user should be aware of - not something temporary to warn folks of. 

I'd suggest either going with the blue informational alert or keeping this as plain text at the beginning of the component documentation. It seems like required documentation core to all the forms components.

colebemis (approved):

emilybrick (approved):

PR #489, opened 2023-05-17
Update disability data

ichelsea:
Good find from @broccolinisoup that the World Health Organization (WHO) has updated their world disability percentage estimate. This PR updates the percentage and the link.


broccolinisoup (approved):
Thank you! 🙇🏻‍♀️

PR #496, opened 2023-05-31
Add Rails migration guides + a few missing components

camertron:
This PR copies over the existing Rails migration guides [here](https://primer.style/view-components/migration). A couple of additional components mentioned in the guide haven't been brought over yet, so I also added them.


camertron:
@colebemis any idea why the preview deploy is failing? It says there's something wrong with the Autocomplete Figma page, but it seems to work fine locally?

PR #502, opened 2023-06-06
Bump doctocat theme to prevent SSRProvider warning messages

camertron:
See: https://github.com/primer/doctocat/pull/586

Seems to unblock deploys 🎉


colebemis (approved):
You're my hero 💖

PR #509, opened 2023-06-10
Update header navigation

colebemis:
- Removes `/ Interface guidelines` from header
- Updates top-level nav links as described in https://github.com/github/primer/issues/2323
- Updates header to say "Primer Design System" instead of "Interfaces Guidelines"

## Before

Before screenshot

## After

After screenshot

Closes https://github.com/github/primer/issues/2323

## Open question

Do we want the top-level nav links to change for all primer.style sites or just primer.style/design? cc @emilybrick

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -2,8 +2,8 @@ const path = require('path')
 
 module.exports = {
   siteMetadata: {
-    title: 'Interface guidelines',
-    shortName: 'Interface guidelines',
+    title: 'Primer',
+    shortName: '',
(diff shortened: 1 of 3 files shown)


gatsby-config.js
```diff
@@ -2,8 +2,8 @@ const path = require('path')
 
 module.exports = {
   siteMetadata: {
-    title: 'Interface guidelines',
-    shortName: 'Interface guidelines',
+    title: 'Primer',
+    shortName: '',
```
colebemis, line 6:
```suggestion
    // Remove short name from site header
    shortName: '',
```

emilybrick (approved):

PR #510, opened 2023-06-12
Update rails.mdx

howdyray:
Fixes broken link path to form design guidance.

### DIFF
--- a/content/ui-patterns/forms/rails.mdx
+++ b/content/ui-patterns/forms/rails.mdx
@@ -11,7 +11,7 @@ export default FormsLayout
 
 Creating accessible forms can be challenging. Forms are a common source of accessibility issues, and can hinder someone from successfully completing and submitting a form. This can be true even for people who do not use assistive technologies.
 
-The Primer forms framework is designed to be accessible by default, integrate deeply with Rails, follow web standards, and adhere to Primer's <Link href="/ui-patterns/forms/overview">form interface guidelines</Link>.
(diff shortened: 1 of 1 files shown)


content/ui-patterns/forms/rails.mdx
```diff
@@ -11,7 +11,7 @@ export default FormsLayout
 
 Creating accessible forms can be challenging. Forms are a common source of accessibility issues, and can hinder someone from successfully completing and submitting a form. This can be true even for people who do not use assistive technologies.
 
-The Primer forms framework is designed to be accessible by default, integrate deeply with Rails, follow web standards, and adhere to Primer's <Link href="/ui-patterns/forms/overview">form interface guidelines</Link>.
+The Primer forms framework is designed to be accessible by default, integrate deeply with Rails, follow web standards, and adhere to Primer's <Link href="/design/ui-patterns/forms/overview">form interface guidelines</Link>.
```
camertron, line 14: This is correct for prod, but will unfortunately not work in dev because `/design` is only prepended in prod :(

camertron (changes requested):

content/ui-patterns/forms/rails.mdx
camertron: I think the fix here is to import `Link` from `gatsby` instead of from `primer/react`. Then it should Just Work™.

```javascript
import {Link} from 'gatsby'
```

emilybrick:
closing - fixed in https://github.com/primer/design/pull/539

PR #512, opened 2023-06-13
Missing link on Form Control

arisacoba:
On the page, [Form Contro](https://primer.style/design/components/form-control)l, the link under the `Forms guidance` and `Forms` return 404. I believe the correct one should have `/overview` appended on the URL?

Which is this one https://primer.style/design/ui-patterns/forms/overview


josepmartins (approved):
Good catch! thank you @arisacoba

PR #513, opened 2023-06-15
Add high-level accessibility guidance for DataTable pagination, loading, and sorting

ericwbailey:
This PR adds high-level accessibility guidance for DataTable pagination, loading, and sorting. 

It is intended to serve as notes for authors to consider/be aware of as concerns, and not an exhaustive breakdown of all the various considerations that come with each aspect of the component.  I was thinking we could link to relevant sections in the DataTable API when it is merged to supply that additional information.

### DIFF
--- a/content/components/data-table.mdx
+++ b/content/components/data-table.mdx
@@ -575,8 +575,7 @@ The table must have a title, and can optionally have a subtitle to describe more
 <Box display="flex" flexDirection={['column', 'column', 'column', 'column', 'row']} sx={{gap: 3}}>
 
 <Box as="p" flex={1}>
-  The row header string will be prepended to the ARIA text of action buttons. For examples, Download {`{row header}`}“,
-  “Actions:{`{row header}`}”
+  The row header string will be prepended to the ARIA text of action buttons. For examples, Download {`{row header}`}“, “Actions:{`{row header}`}”
 </Box>
 
 <img
@@ -589,9 +588,7 @@ The table must have a title, and can optionally have a subtitle to describe more
 <Box display="flex" flexDirection={['column', 'column', 'column', 'column', 'row']} sx={{gap: 3}}>
 
 <Box as="p" flex={1}>
-  To handle cases where the row header can be very long, we should give consumers the option to specify a shorter string
-  to identify the rows. For example, if an issue title is used as the row header, we could use the issue number instead
-  of the full issue title.
+  To handle cases where the row header can be very long, we should give consumers the option to specify a shorter string to identify the rows. For example, if an issue title is used as the row header, we could use the issue number instead of the full issue title.
 </Box>
 
 <img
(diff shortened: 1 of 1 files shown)


content/components/data-table.mdx
```diff
@@ -600,6 +600,20 @@ The table must have a title, and can optionally have a subtitle to describe more
   src="https://user-images.githubusercontent.com/2313998/205361249-763772e1-d856-479e-a632-d185c27095db.png"
 />
 </Box>
+
+### Pagination buttons
+
+<Box as="p" flex={1}>
+  The pagination control for a DataTable uses `button` elements, and not `a` elements. This is because activating the
```
mperrotti, line 607: @joshblack - just want to confirm that DataTable pagination can _only_ ever be buttons, and there's no way for consumers to render links instead.

content/components/data-table.mdx
```diff
@@ -600,6 +600,20 @@ The table must have a title, and can optionally have a subtitle to describe more
   src="https://user-images.githubusercontent.com/2313998/205361249-763772e1-d856-479e-a632-d185c27095db.png"
 />
 </Box>
+
+### Pagination buttons
+
+<Box as="p" flex={1}>
```
mperrotti, line 606: Why do we need `<Box as="p" flex={1}>`? I think this would render just fine as regular markup.

content/components/data-table.mdx
```diff
@@ -600,6 +600,20 @@ The table must have a title, and can optionally have a subtitle to describe more
...
+  The pagination control for a DataTable uses `button` elements, and not `a` elements. This is because activating the 
+  pagination control requests data, and does not have a corresponding URL to link to.
```
mperrotti, line 608: We shouldn't have line breaks in `.mdx` text blocks unless they're used to add a breaking space or create a new paragraph.

content/components/data-table.mdx
```diff
@@ -600,6 +600,20 @@ The table must have a title, and can optionally have a subtitle to describe more
...
+</Box>
+
+### Loading and sorting content
+
+<Box as="p" flex={1}>
+  It is important to ensure parity between what can be seen visually when a table is loading or sorting content and 
+  what is announced to assistive technology. This is provided by the DataTable component automatically, but must be considered if a custom loading solution is otherwise utilized. For loading content, this also includes the initial loading period.
```
mperrotti, line 615: I think we should describe the "batched" loading announcements.

mperrotti, line 615: > This is provided by the DataTable component automatically

I think only the announcement for sorting is provided by the DataTable component automatically. I think the loading announcement(s) have to be handled by consumers. @joshblack - please correct me if I'm wrong.

mperrotti (commented):
Looking good! The only "blocking" feedback is that I think we should describe the "batched" loading announcements in more detail.

content/components/data-table.mdx
joshblack: Yup! Only buttons for now.

content/components/data-table.mdx
joshblack: Yeah, this may be best left up to consumers but I think if we can find some common patterns we can try to codify this in the same way that TreeView does either through a prop (like `status`) or through things returning promises that either resolve (success) or reject (error) and potentially handle things through there 🤷 Definitely an open question for me still and am happy to explore this to see what makes it easy on folks using this component

joshblack (approved):

content/components/data-table.mdx
mperrotti: Note: this only applies if we're not wrapping the text in a `<p>` or `<Box as="p">`

content/components/data-table.mdx
ericwbailey: For posterity: Made a note to pitch this as a Primer consideration.

PR #519, opened 2023-06-27
Fix links generated by the rails-markdown component

camertron:
Another casualty of the recent change that added statuses to URLs 🤦


keithamus (approved):
:shipit:

PR #524, opened 2023-07-04
Fix broken link

Daeraxa:
Link to `repo` icon page is broken on https://primer.style/design/foundations/icons/design-guidelines#corners as it seems to be looking for it on the root and not relative to the icons page.

Currently leads to https://primer.style/design/repo-16/ (a 404).


keithamus (approved):
Thanks for this!

PR #528, opened 2023-07-12
Fix React component fetching

camertron:
We've received reports of errors when starting up the Gatsby dev server. It looks like the generated/ directory [was recently removed](https://github.com/primer/react/pull/3479) from primer/react and is now treated as a build artifact. Gatsby is attempting to fetch this file from the repo, which fails. This PR fixes things by fetching generated/components.json from unpkg instead of from the primer/react git repo directly.


maximedegreve (approved):
Merge this in my branch and this fix worked! 🚀

PR #532, opened 2023-07-18
About page layout

josepmartins:
`About` page layout and style changes 

Fixes: https://github.com/github/primer/issues/2425

👀 https://primer-b621672794-26441320.drafts.github.io/about

- [x] Swap /about to a full-width layout
- [x] Fix margin underneath the h1
- [x] Remove "About" from the sidebar

### DIFF
--- a/content/about.mdx
+++ b/content/about.mdx
@@ -2,7 +2,7 @@
 title: Meet the team
 ---
 
-import {Team} from '../../src/components/team'
+import {Team} from '~/src/components/team'
 import MinimalLayout from '~/src/layouts/minimal-layout'
 export default MinimalLayout
 
--- a/content/guides/about.mdx
+++ b/content/guides/about.mdx
@@ -3,5 +3,7 @@ title: Meet the team
 ---
 
 import {Team} from '../../src/components/team'
+import MinimalLayout from '~/src/layouts/minimal-layout'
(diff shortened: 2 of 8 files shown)


mperrotti (commented):
I like the other version better. The bios bleed together a little bit in this one, and it's not as "fun" or expressive as the other design.

mperrotti:
Odd even layout looks good! I'd be happy with that.

emilybrick (approved):
✨ 💖 Have two small requests, but I don't want to block you tm morning - 

Can we increase the margin underneath the "Meet the team" header? It feels a tad tight on top of the description
Screenshot 2023-07-19 at 2 59 18 PM

Separately - the new URL will be primer.style/about, right? I think this was the case, just making sure 🔗

josepmartins:
> Can we increase the margin underneath the "Meet the team" header? It feels a tad tight on top of the description

I'm ok adding more margin, I was following the latest layout updates. Should we add it to the rest of layouts too? 

Example for non-component layouts: https://primer-5401c81006-26441320.drafts.github.io/guides/introduction


> Separately - the new URL will be primer.style/about, right? I think this was the case, just making sure

Shouldn't the url be `primer.style/design/about` until we make the redirects in this repo?

mperrotti (approved):

PR #541, opened 2023-07-26
Remove System Arguments page since it's being moved to a Lookbook page

mperrotti:
Relates to https://github.com/github/primer/issues/2492

Should not be merged until https://github.com/primer/view_components/pull/2159 is merged.


camertron (approved):
Awesome, thank you!

PR #546, opened 2023-08-01
Fix link to system arguments

camertron:
The link to Rails system arguments is currently broken. This should fix it.


mperrotti (commented):
This looks correct, but I think we may also need to update the link in `nav.yml`

https://github.com/primer/view_components/blob/main/docs/src/%40primer/gatsby-theme-doctocat/nav.yml

camertron:
See: https://github.com/primer/view_components/pull/2184

PR #551, opened 2023-08-09
Add CSS documentation, deprecated section to components

emilybrick:
This PR adds a CSS tab to the existing Primer components that have them. They are technically in a section in the CSS Storybook called "deprecated", and each CSS tab should have clear messaging that these components are not being actively worked on. 

**Notes:** 
- ~I'm pulling in the [entire embed](https://primer.style/css/storybook/?path=/docs/deprecated-branchname--docs) with docs, unsure how to break out the documentation of the example and have it just appear in the MDX (I don't _think_ we do this anywhere, see below). I could just use the [default](https://primer.style/css/storybook/?path=/story/deprecated-branchname--default) example in Storybook instead, but would have no corresponding docs~  I'm just pulling in the [default](https://primer.style/css/storybook/?path=/story/deprecated-branchname--default) component story example, without docs
- Can't get the `Show code` button to appear, like it does w/ React embeds?

Screenshot 2023-08-09 at 11 56 23 AM

Notes + Follow ups:

- This adds a deprecated section for components we no longer encourage the use of.
- In a follow up, we should add a "View code" button on the CSS storybook embeds, which requires generating a components.json file to include the code used in the stories from the CSS repo


src/layouts/rails-component-layout.tsx
```diff
@@ -29,6 +29,7 @@ export const query = graphql`
           reactId
           railsIds
           figmaId
+          cssId
```
emilybrick, line 32: I'm not sure what I'm missing but the CSS tab is not appearing if you're current page is the React, Rails, or Figma tab. The CSS tab only shows up if you're on the current tab, or if you're looking at the Overview

src/layouts/css-component-layout.tsx
```diff
@@ -0,0 +1,118 @@
...
+                <SourceLink href={`https://github.com/primer/css/blob/main/src/${name}`} />
+                {stories.length > 0 ? (
+                  <StorybookLink href={`https://primer.style/css/storybook/?path=/story/${stories[0].id}`} />
+                ) : null}
+              </Box>
+            </Box>
+
+            <H2>Examples</H2>
+            {stories.length > 0 ? (
+              <StorybookEmbed framework="css" height={300} stories={stories} />
```
emilybrick, line 100: can't seem to get the `Show Code` button to appear on the embed...is that bc of CSS Storybook configuration?

src/layouts/rails-component-layout.tsx
josepmartins: you might need to add the CSS `Undelinenav.Link` to the the Rails/React and Figma layouts like you did here for Overview:

https://github.com/primer/design/pull/551/files#diff-c4827c6a111be693621e9ce1a0056a1e714d324742542bb266cdae9045a52679R40-R44

src/layouts/css-component-layout.tsx
josepmartins: It's not Storybook related. It seems the `code` data is passed and generated on the React side and it's part of the component.json file: https://unpkg.com/@primer/react@35.27.1/generated/components.json

React code docs generation: https://github.com/primer/react/blob/main/script/components-json/build.ts#L119

If we want the `Show Code` we'll need to generate a similar components.json file including the code used in the stories in the CSS repo 😄

emilybrick: This seems like a good chunk or work - let's handle in a follow-up PR?

mperrotti: I think it's ok to do this in a follow-up PR

mperrotti (commented):
I'm not seeing a CSS tab

Screenshot 2023-08-17 at 2 33 15 PM

emilybrick:
> I'm not seeing a CSS tab
> 
> Screenshot 2023-08-17 at 2 33 15 PM

This is odd. I'm on same preview deploy link:
Screenshot 2023-08-18 at 10 24 26 AM

mperrotti (approved):
I'm seeing a CSS tab after clearing cache

PR #556, opened 2023-08-11
Add mistakenly deleted .tooltipped migration docs

camertron:
These docs were deleted when we removed the old docsite 😳


jonrohan (approved):
Thanks 🤘🏻

PR #557, opened 2023-08-11
Expose dotcom shared components

camertron:
Partially addresses https://github.com/github/primer/issues/2310

This PR adds the list of shared React components from dotcom:

A screenshot of a page on primer.style showing a list of component name links grouped alphabetically by first letter. Two components are listed: ItemPicker and ListView. An icon to the right of each component indicates clicking on the link will open a new tab/window. A label next to the icon says 'draft.'

The list of shared components comes from a build artifact in dotcom CI. It is downloaded by primer/design's CI and used to populate the list on primer.style.

Note that not all components in dotcom have corresponding stories. For those components that do not have stories, the links navigate to the component's source code on github.com.

Last thing of note: we wanted the shared components to appear in search results. I accomplished this by making a [change](https://github.com/primer/doctocat/pull/624) to the doctocat template that allows adding arbitrary data to the search corpus. Each shared component needs to have its own URL for this to work. I've added redirects for each of these faux pages that takes you back to the list of shared components.


content/github-staff/github-shared-components.mdx
```diff
@@ -21,3 +24,5 @@ Please note that these components are **owned by the feature teams**, and will n
 
 - [Handling new patterns](/guides/contribute/handling-new-patterns)
 - [Upstreaming to Primer](/guides/contribute/adding-new-components#upstreaming-to-primer)
+
+## Components
```
emilybrick, line 28:
```suggestion
## Shared components
```

gatsby-node.js
```diff
@@ -32,6 +33,55 @@ exports.sourceNodes = async ({actions, createNodeId, createContentDigest}) => {
   await sourcePrimerRailsData({actions, createNodeId, createContentDigest})
   await sourceOcticonData({actions, createNodeId, createContentDigest})
   await sourceFigmaData({actions, createNodeId, createContentDigest})
+  await sourceDotcomSharedComponentsData({actions, createNodeId, createContentDigest})
+}
+
+async function sourceDotcomSharedComponentsData({actions, createNodeId, createContentDigest}) {
+  // @TODO: eventually get this from a dotcom build artifact
+  const sharedComponents = JSON.parse(fs.readFileSync('shared_components.json'))
```
mperrotti, line 41: Is this work being planned? I'm worried about this becoming quickly outdated and not trusted.

content/github-staff/github-shared-components.mdx
```diff
@@ -21,3 +24,5 @@ Please note that these components are **owned by the feature teams**, and will n
 
 - [Handling new patterns](/guides/contribute/handling-new-patterns)
 - [Upstreaming to Primer](/guides/contribute/adding-new-components#upstreaming-to-primer)
+
+## Shared components
```
mperrotti, line 28: Now that we have this section, I think the "Finding these components" section is redundant. Let's move the content from that section into the new one.

mperrotti, line 28: The links in this section don't match Primer's link styles.

Screenshot 2023-08-15 at 5 40 55 PM

emilybrick:
Sharing some feedback from @dipree, who is working on organizing the internal Storybook for these components, and the concept of Recipes 👇 

Explanation from the docs of how recipes fit in:

- **Primer Components** - Design System Components are shared, content-agnostic, and context-agnostic built for maximal reuse. These components leverage Primer Primitives and have a very high standard. Primer has contribution guidelines to help achieve this level of quality.
- **Primer Recipes** - Recipe components are mainly compositions of design system components that are to be used consistently in GitHub. Most often, they are not agnostic enough to live in the design system therefore, they are not public on primer.style. Sometimes they can become a Primer component. Recipes can be created by any team at GitHub and are maintained by the team that created them.
- **One-offs** - In rare cases, one-off components are needed for very specific use cases. But for the most part, creating them should be avoided. Custom solutions can be great but run the risk of causing an inconsistent user experience leading to customer friction.


Right now, the `Shared components` list displayed in **this PR** align most closely with the `One-offs` described in the bullets above ☝️ 

From @dipree:
> I would say we only surface Recipes and Recipes [Draft] in Primer. The rest is considered “experimental”.

To see an example of Recipes in the storybook (WIP): https://github.com/github/github/pull/283177

/cc @camertron @mperrotti

gatsby-node.js
camertron: Yes, I'm currently working on it in parallel.

camertron:
@emilybrick cool, thanks for the recipes context. Should we make it clear the shared components we're linking to are experimental then, or only list recipes? I think there are only like 2 recipes at the moment.

lesliecdubs:
@camertron let's see what Emily says as well, but think the updated direction here would be to _only_ expose Recipes in the Primer docs, and not list the experiments at all. We're in early stages of Recipes so it's okay if there's only those first 2 to start.

emilybrick (commented):
Thank you for updating this so quickly!

Two comments/questions but overall this LGTM

- I don't think we need the alphabetical headings at this point, since there are only two components - could just be bulleted list?
- do we need to say these are "recipes" anywhere? /cc @lesliecdubs @dipree I'm fine calling them components as its a bit more straightforward and introducing the recipe concept on the public site might not be necessary.

lesliecdubs:
> do we need to say these are "recipes" anywhere? /cc @lesliecdubs @dipree I'm fine calling them components as its a bit more straightforward and introducing the recipe concept on the public site might not be necessary.

My two cents: I do think using the term “recipe” would be helpful for Hubbers to make the connection. That said, I’m in favor of adding/explaining the recipe concept on this page in an iteration once Daniel is back so we can go ahead and get this addition out in the meantime.

emilybrick (approved):

PR #558, opened 2023-08-14
Fix component status table

camertron:
The [component status table](https://primer.style/design/guides/status) recently broke because we [completely removed](https://github.com/primer/view_components/pull/2193) the legacy PVC docsite, including the necessary components.json file it published. This PR introduces a `RailsProvider` React provider that includes refactored versions of helper methods that associate Rails and React components together.


rezrah (approved):
Works great 🎉, thanks @camertron

PR #565, opened 2023-08-24
Update `Tooltip` visual design in interface guidelines + accessibility guidelines

langermank:
tooltip anatomy graphic

## Goals

- Update the visual design of the Tooltip component (remove caret)
- Surface the tooltip alternatives into the main interface guidelines to improve discoverability
- Add a few more usage guidelines and explain label vs. description

Accessibility page → https://primer-14e523666e-26441320.drafts.github.io/guides/accessibility/tooltip-alternatives
Interface guidelines → https://primer-14e523666e-26441320.drafts.github.io/components/tooltip

**TODO**

- [ ] After merging, followup and update these links: https://github.com/primer/design/pull/565/files#r1305710242


content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+>
+  <img
+    width="456"
+    alt="icon button with tooltip as a description"
+    src="https://github.com/primer/design/assets/18661030/2d0cd085-2454-406e-8c99-bc8bd3dd20db"
+  />
+  <Box as="p" mt="0">
+    For interactive controls that require additional context, a tooltip can be used to provide it. The `label` is also
+    required, and the description will be announced to a screen reader in addition to the label. Keep the description
+    short and descriptive.
```
langermank, line 119:
```suggestion
    For interactive controls that require additional context, a tooltip can be used to provide it. The `label` is also
    required, and the description will be announced to a screen reader in addition to the label. Keep the description
    short and concise.
```

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+By default, the Tooltip component will position itself below the interactive control in a centered alignment. The position is dependant on the space available within the viewport, and it will never break outside of it. It may appear to the top, bottom, left, or right of the interactive control.
+
+<img
+  width="960"
+  role="presentation"
+  src="https://github.com/primer/design/assets/18661030/c620d282-1efd-4556-9d9c-770c37359fc9"
+  alt="tooltips rendered in all possible positions surrounding a control"
+/>
+
+## Alternatives to tooltips
```
mperrotti, line 134: Love this section so much

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+    <img
+      width="456"
+      alt="Non-interactive text element activating a long and wordy tooltip on hover"
+      src="https://github.com/primer/primitives/assets/18661030/1d8d86a2-8ee7-40d2-8934-c3d36f34d9b3"
+    />
+    <Caption>Don't use tooltips on non-interactive elements</Caption>
+  </Dont>
+</DoDontContainer>
+
+<DoDontContainer>
```
mperrotti, line 58: I don't understand the intention of this guideline, but I'm okay with it.

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+  mb={3}
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column', 'column', 'column', 'column', 'row']}
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="icon button with tooltip as a description"
+    src="https://github.com/primer/design/assets/18661030/2d0cd085-2454-406e-8c99-bc8bd3dd20db"
```
mperrotti, line 114: I think we try and avoid `aria-label`, but this might be ok. cc @ericwbailey - what do you think?

mperrotti (approved):

content/components/tooltip.mdx
khiga8, line 134: We have a couple of lint rules and doc pages where we cross-reference the existing [Tooltip alternatives](https://primer.style/design/guides/accessibility/tooltip-alternatives) docs. Let's make sure to update those to avoid a dead link! (I can verify tomorrow which ones those are). 

Or could it be sufficient to just cross-reference that page from this section rather than moving it entirely?

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+    <Caption>Don't use tooltips on non-interactive elements</Caption>
+  </Dont>
+</DoDontContainer>
+
+### Other
+
+If you are unsure which alternative is more suited to your scenario and need help, consult a designer or the GitHub Accessibility team (if you are GitHub staff) for advice.
 
-Most UI cases don't call for tooltips. Learn some [alternative methods to use in place of tooltips](/../guides/accessibility/tooltip-alternatives).
+## Additional resources
```
khiga8, line 238: We [introduced a new PVC lint rule](https://github.com/primer/view_components/pull/2086) recently the deprecated Primer CSS tooltipped which we have a ton of debt around that we want to tackle. It looks like the lint rule doc is now on [Primer CSS Tooltipped](https://primer.style/design/guides/development/rails/migration-guides/primer-css-tooltipped).
Could it make sense to link to it here? :)

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+    <Caption>
+      Don’t use tooltips for icon buttons that appear in a group of other controls which share the same functionality,
+      such as opening a context menu.
+    </Caption>
+  </Dont>
+</DoDontContainer>
+
+## Options
+
+### Label
```
khiga8, line 82: Could it be useful to talk about the underlying markup of the differences between a `label` and `description` (e.g. `aria-labelledby` vs `aria-describedby`)? One will serve as the accessible name of a control, while the other will supplement it.

We have some more documentation in [Primer Rails tooltip](https://primer.style/design/components/tooltip/rails/alpha) (This was originally under an accessibility section in Primer Rails but I think it changed with the recent docs site update!)

khiga8 (commented):
First pass! Thank you for working on this :) 
I hope to take a closer look tomorrow!

content/components/tooltip.mdx
langermank: Yes, thank you! I had a hard time trying to explain this one so let me see if I can rework it a bit.

lukasoppermann (approved):
😍

lukasoppermann:
Hey @langermank could you add a Figma section as well?

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
   role="presentation"
-  src="https://github.com/primer/design/assets/586552/787873b7-d496-454c-b669-0c2400ef503f"
+  src="https://github.com/primer/primitives/assets/18661030/172aa8f6-2273-4719-bb63-8dd678ffcbc9"
   alt="three examples of tooltips on interactive elements"
 />
 
-## Accessibility
+- Reserve tooltips for components such as icon buttons.
+- Keep the tooltip text minimal.
+- **Never** include tooltips on non-interactive components (`div`, `span`, `p`).
```
khiga8, line 35:
```suggestion
- **Never** include tooltips on non-interactive components (`div`, `span`, `p`) because it will not be accessible for keyboard or screen reader users.
```

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+### Best practices
+
+<DoDontContainer>
+  <Do>
+    <img
+      width="456"
+      alt="Icon buttons with tooltips for edit, favorite, and sponsor"
+      src="https://github.com/primer/primitives/assets/18661030/e9f86af8-3792-4794-af68-ad0efccd8bc8"
+    />
+    <Caption>Use a tooltip to give a text label to an icon button</Caption>
```
khiga8, line 46: Could it be useful to mention why this is helpful for an icon button? (e.g. to visually convey the purpose of a button since icons are not always recognizable)

content/components/tooltip.mdx
khiga8: * [Primer CSS Tooltipped](https://primer.style/design/guides/development/rails/migration-guides/primer-css-tooltipped)
* [a11y-role-supports-aria-props](https://github.com/github/eslint-plugin-github/blob/b93f04ae4600c18f86d65e4d5854820f71ed26bb/docs/rules/a11y-role-supports-aria-props.md?plain=1#L16)
* [NoAriaLabelMisuse](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/no-aria-label-misuse.md?plain=1#L10)
* [a11y-no-title-attribute](https://github.com/github/eslint-plugin-github/blob/b93f04ae4600c18f86d65e4d5854820f71ed26bb/docs/rules/a11y-no-title-attribute.md?plain=1#L11)
* [a11y_no_rogue_role_equals_tooltip](https://github.com/github/github/blob/5dc32d30d7d08078c30df315fd2b6f750abc6ba8/.erb-linters/accessibility/a11y_no_rogue_role_equals_tooltip.rb#L15)
* [PVC Tooltip](https://github.com/primer/view_components/blob/main/app/components/primer/alpha/tooltip.rb#L6)

content/components/tooltip.mdx
khiga8: I don't think this is shipped in React. Would we want to hold off on this section to avoid confusing people 🤔

(cc: @broccolinisoup who worked on React tooltip :) )

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+<Box
+  mb={3}
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column', 'column', 'column', 'column', 'row']}
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="icon button with tooltip as a description"
```
khiga8, line 113:
```suggestion
    alt="Icon button with tooltip as a description"
```

content/components/tooltip.mdx
```diff
@@ -3,42 +3,239 @@ title: Tooltip
...
+    src="https://github.com/primer/design/assets/18661030/2d0cd085-2454-406e-8c99-bc8bd3dd20db"
+  />
+  <Box as="p" mt="0">
+    For interactive controls that require additional context, a tooltip can be used to provide it. The `label` is also
+    required, and the description will be announced to a screen reader in addition to the label. Keep the description
```
khiga8, line 118: I think we will want to reword this a bit since this seems to suggest one should set both the `label` and `description` prop of the tooltip API which I don't think is the case. Maybe we want to say that when a tooltip supplements a control as a description, we should make sure the control has an accessible name?

```
<button aria-describedby="tooltip-1">Name of button</button>
<tooltip id="tooltip-1">Some supplementary description</tooltip>
```

content/components/tooltip.mdx
khiga8, line 82: Could we mention how a label type tooltip should be reserved for icon buttons? The underlying semantic of a label tooltip is `aria-labelledby` would override any label the button already has.

content/components/tooltip.mdx
broccolinisoup: ➕ 1 - Would love to learn! It might be good idea to briefly explain the reasoning as well

content/components/tooltip.mdx
broccolinisoup: >Yes, thank you! I had a hard time trying to explain this one so let me see if I can rework it a bit.

Katie let me know if you want to pair on this or any help with it!

>I don't think this is shipped in React. Would we want to hold off on this section to avoid confusing people

Yes! We are going to ship the draft version soon hopefully 🙏🏻  and also the visual changes (i.e. removing the caret) also will be shipped with the draft version so holding off merging this PR until the draft is out would be wise.

content/components/tooltip.mdx
broccolinisoup: >we should make sure the control has an accessible name?

I agree with this. Since accessible name could be a visible button text or an `aria-label`, we can maybe divide the "Description" section into two subsections?

Description for buttons that has a visible name
```html
<button aria-describedby="tooltip-1">Name of button</button>
<tooltip id="tooltip-1">Some supplementary description</tooltip>
```

Description for buttons that doesn't have a visible name
```html
<iconbutton aria-label="icon-button-label" />
<tooltip id="tooltip-1">Some supplementary description</tooltip>
```

broccolinisoup (commented):
So cool 🔥  Left some comments, let me know what you think 🙌🏻

content/components/tooltip.mdx
langermank: I decided to remove this section, and instead we can update IconButton docs to talk about when to _not_ use a tooltip

content/components/tooltip.mdx
langermank: I feel like this is clear already by saying "For interactive controls with no visible text label" but I will add "such as icon buttons" 😄

content/components/tooltip.mdx
langermank: I pulled this from the markup that I think has gone through all the a11y steps, so this is how it is under the hood. I'm trying to show that this IconButton has a description in addition to its aria-label.

content/components/tooltip.mdx
langermank: I'll follow up with the links outside this repo once this PR merges! ❤️ thank you

PR #583, opened 2023-09-05
Add nested dialog policy

khiga8:
Fixes: https://github.com/github/accessibility/issues/4331

Codifying [past office hours guidance from James](https://github.rewatch.com/video/3grwgizxhzn0u6ab-accessibility-design-office-hours-october-4-2022?t=1014). 

Anything else we should document here?

cc: @primer/accessibility-design

### DIFF
--- a/content/components/dialog.mdx
+++ b/content/components/dialog.mdx
@@ -5,7 +5,7 @@ description:
   more.
 reactId: dialog
 railsIds:
-- Primer::Alpha::Dialog
+  - Primer::Alpha::Dialog
 figmaId: dialog
 rails: https://primer.style/view-components/components/alpha/dialog
 ---
@@ -255,6 +255,12 @@ If the answer is yes, designing the feature inside a dialog provides an easy way
   </Dont>
 </DoDontContainer>
 
+### Nested dialogs
+
+It is acceptable to have up to two nested dialogs as long as the dialog is implemented correctly. For example, upon closing the second nested dialog, focus should return to the trigger control for this dialog within the first dialog. When the two stacked dialogs are open, pressing `Esc` should only close the top-level dialog.
+
+Howeve, avoid having more than three nested dialogs as this adds unnecessary complexity to the user experience.
+
 ### Deep linking
 
 It's crucial to consider if the dialog should be accessible via a link. For instance, if our users intend to link to this specific content from outside GitHub.
@@ -346,7 +352,9 @@ This consideration is essential for maximizing productivity.
   </Do>
   <Dont>
     <img width="464" alt="" src="https://github.com/primer/design/assets/980622/df59604f-c459-4aff-8db7-3e23000c91ac" />
-    <Caption>Unless it's a global action, avoid side sheets in pages where content does not stretch the full width of the page.</Caption>
+    <Caption>
+      Unless it's a global action, avoid side sheets in pages where content does not stretch the full width of the page.
+    </Caption>
   </Dont>
 </DoDontContainer>
 
--- a/content/components/dialog.mdx
+++ b/content/components/dialog.mdx
@@ -257,7 +257,7 @@ If the answer is yes, designing the feature inside a dialog provides an easy way
 
 ### Nested dialogs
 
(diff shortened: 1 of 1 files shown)


content/components/dialog.mdx
```diff
@@ -5,7 +5,7 @@ description:
   more.
 reactId: dialog
 railsIds:
-- Primer::Alpha::Dialog
+  - Primer::Alpha::Dialog
```
khiga8, line 8: autoformatted

content/components/dialog.mdx
```diff
@@ -346,7 +352,9 @@ This consideration is essential for maximizing productivity.
   </Do>
   <Dont>
     <img width="464" alt="" src="https://github.com/primer/design/assets/980622/df59604f-c459-4aff-8db7-3e23000c91ac" />
-    <Caption>Unless it's a global action, avoid side sheets in pages where content does not stretch the full width of the page.</Caption>
+    <Caption>
```
khiga8, line 355: autoformatted

content/components/dialog.mdx
```diff
@@ -255,6 +255,12 @@ If the answer is yes, designing the feature inside a dialog provides an easy way
   </Dont>
 </DoDontContainer>
 
+### Nested dialogs
+
+It is acceptable to have up to two nested dialogs as long as the dialog is implemented correctly. For example, upon closing the second nested dialog, focus should return to the trigger control for this dialog within the first dialog. When the two stacked dialogs are open, pressing `Esc` should only close the top-level dialog.
```
owenniblock, line 260: Add a note here about clicking outside the second dialog only closing the second dialog as well?

content/components/dialog.mdx
```diff
@@ -255,6 +255,12 @@ If the answer is yes, designing the feature inside a dialog provides an easy way
...
+
+However, having more than three nested dialogs is discouraged because it adds unnecessary complexity to the user experience.
```
owenniblock, line 262: Non-blocking but thought it might be useful to expand on this slightly.

```suggestion
However, having more than three nested dialogs is discouraged because it adds unnecessary complexity to the user experience and can be disorienting for some users.
```

owenniblock (approved):
Looks good to me, couple of thoughts but I'd be happy with it like this 🤷 😄

content/components/dialog.mdx
khiga8, line 260:
```suggestion
It is acceptable to have up to two nested dialogs as long as the dialog is implemented correctly. For example, upon closing the second nested dialog, focus should return to the trigger control for this dialog within the first dialog. When the two stacked dialogs are open, pressing `Esc` should only close the top-level dialog. When one clicks outside of the second dialog with a mouse, only the second dialog should close.
```

content/components/dialog.mdx
```diff
@@ -255,6 +255,12 @@ If the answer is yes, designing the feature inside a dialog provides an easy way
   </Dont>
 </DoDontContainer>
 
+### Nested dialogs
+
+It is acceptable to have up to two nested dialogs as long as the dialog is implemented correctly. For example, upon closing the second nested dialog, focus should return to the trigger control for this dialog within the first dialog. When the two stacked dialogs are open, pressing `Esc` should only close the top-level dialog. When one clicks outside of the second dialog with a mouse, only the second dialog should close.
```
JoyceZhu, line 260: I think it'd be helpful to link "implemented correctly" to some actionable guidelines: for example, the [ARIA spec](https://www.w3.org/TR/wai-aria-1.2/#dialog) for the `dialog` role or the [Primer Dialog component](https://primer.style/react/Dialog)

khiga8: The Primer React Dialog hasn't been accessibility remediated yet. The ARIA spec doesn't have as much guidance as I'd hoped. Maybe https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/?

khiga8: It's not nested dialog specific guidance though. I feel okay leaving out given we have written out the behavior we're expecting for nested dialogs.

khiga8: This page has a [Trigger elements and keyboard focus](https://primer.style/design/components/dialog#trigger-elements-and-keyboard-focus) section which describes expected behavior. I will link to this.

khiga8: What do you think of https://github.com/primer/design/pull/583/commits/03a3bdf6beeb31fd6e75a8eab920c2052a2f4621?

JoyceZhu (approved):

TylerJDev (approved):
Very nice! Thanks for adding this guidance, looks great! ✨

Issue #572, opened 2023-08-25
Broken image links in icon design guidelines


gavinmn:
It seems like the image links in https://github.com/primer/design/blob/main/content/foundations/icons/design-guidelines.mdx are no longer working

![Image](https://github.com/primer/design/assets/59900904/d36321c1-6b0a-4e1c-b4aa-c861eed667c8)

josepmartins:
Hey @gavinmn thanks for the heads up. Adding more context to the issue, the URLs in src appear to be broken as they’re not hosted on GitHub, they’re hosted on https://compai.pub/, which appears to be a broken Netlify site. We need to re-upload them.

Slack thread: https://github.slack.com/archives/C03L4CK58BC/p1693232293826019

We'll take a look at it as soon as possible 👍

emilybrick:
For posterity, Figma source images are here: https://www.figma.com/file/2rjsQp42uKvzPXzLZIqOeA/Octicons-v2---Source?type=design&node-id=3064%3A1807&mode=design&t=vhZUsuTQweKsx2k7-1

josepmartins:
Closing as the issue was fixed here: https://github.com/primer/design/pull/585

PR #585, opened 2023-09-06
Icon design guidelines missing images

josepmartins:
Fixes broken image links in icon design guidelines.

Tracking issue: https://github.com/primer/design/issues/572

### DIFF
--- a/content/foundations/icons/design-guidelines.mdx
+++ b/content/foundations/icons/design-guidelines.mdx
@@ -8,21 +8,29 @@ These are guidelines to use as a reference when designing new octicons.
 
 ## Grid sizes
 
-Always design two versions of each icon: a 16px version and a 24px version. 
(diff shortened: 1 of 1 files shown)


emilybrick (changes requested):
Hey this looks great - thank you for doing this! 

Small nit - can we fix the single image rows (I think there are only a couple like this) here so they are left aligned with the caption?
Screenshot 2023-09-06 at 11 16 38 AM

josepmartins:
> Hey this looks great - thank you for doing this!
> 
> Small nit - can we fix the single image rows (I think there are only a couple like this) here so they are left aligned with the caption and text above them? Screenshot 2023-09-06 at 11 16 38 AM

Sure! it's now done

emilybrick (approved):

PR #588, opened 2023-09-10
Add info about required css loading

DDKatch:
I have tried to start a new project with primer view components but I faced several issues with it.
You can check my setup here https://github.com/DDKatch/github_primer_components_css_fix
So I have updated the docs with the fix that helped me.

### DIFF
--- a/content/guides/development/rails/index.mdx
+++ b/content/guides/development/rails/index.mdx
@@ -23,17 +23,20 @@ require "view_component"
 require "primer/view_components/engine"
 ```
 
-It's important to load css directly from the gem. Otherwise lots of components will have a "broken css" look.
+##### CSS
 
-Load it as js module into the pipeline. Update app/assets/config/manifest.js with:
+Load CSS as a JS module into the pipeline. Update `app/assets/config/manifest.js` with:
+
 ```js
 //= link primer_view_components.css
 ```
 
-so you can use this inside the template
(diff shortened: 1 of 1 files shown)


content/guides/development/rails/index.mdx
```diff
@@ -23,6 +23,18 @@ require "view_component"
...
+Load it as js module into the pipeline. Update app/assets/config/manifest.js with:
+```js
+//= link primer_view_components.css
+```
+
+so you can use this inside the template
+```erb
+<%= stylesheet_link_tag("primer_view_components") %>
+```
+
```
langermank, line 37: Thanks for this addition! I'm suggesting a slight rewording:


```suggestion
**CSS:**

Load CSS as a JS module into the pipeline. Update app/assets/config/manifest.js with:

```js
//= link primer_view_components.css
```

Import the stylesheet into templates:

```erb
<%= stylesheet_link_tag("primer_view_components") %>
```

```

DDKatch: @langermank thank you for your comments.
I have applied your suggestions and added a bunch of small things, so now the doc looks like this
Screenshot 2023-09-12 at 11 20 12

langermank: Wonderful, looks great! Thank you 🙌

langermank (approved):

PR #591, opened 2023-09-12
Primer Design data visualization guidance

tbenning:
I wrote up some light guidance on data visualizations. Here's the PR! cc: @mperrotti 

https://github.com/primer/design/assets/7265547/d07fc949-2176-4431-bf7d-1536b2b74d45

**Discussion: Presentational colors**
For the section [choosing chart colors](http://localhost:8000/ui-patterns/data-visualization#our-set-of-color-values), we are referencing `scale` values, which are absolute in each color mode and don't scale effectively. This brings up the concept of `presentational` color values again, which could look like Radix's colors, which map to dark mode more effectively. 

e.g. 
image

### DIFF
--- a/content/ui-patterns/charts.mdx
+++ b/content/ui-patterns/charts.mdx
@@ -1,0 +1,65 @@
+---
+title: Data visualization and charts
+description: Data visualizations are helpful tools for conveying complex data in an engaging and understandable way. They're commonly used in dashboards and insights pages.
+---
+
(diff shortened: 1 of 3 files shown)


mperrotti (approved):
These are great!

lukasoppermann:
Love it @tbenning it still is not done due to stuff that got in the way, but I am still working on presentational colors. (Very similar to the redix example.)

PR #592, opened 2023-09-15
Only show highest-ranking status for Rails components on status page

camertron:
At the moment the component status page says the Rails `Button` component has been deprecated 😱 Obviously that's not true! The problem is that we have multiple versions of `Button`. The deprecated one is taking precedence when it shouldn't.


langermank (approved):
Nice!

PR #603, opened 2023-09-20
Fix 'require' in Rails setup docs

camertron:
Apps should require `primer/view_components` instead of `primer/view_components/engine`.


lesliecdubs (approved):
Thanks for this!

PR #608, opened 2023-09-25
Lookbook embeds

camertron:
See also: https://github.com/primer/view_components/pull/2257

This PR uses Lookbook's awesome embedding feature instead of our custom embedding. There are many benefits to using the native embeds, but the biggest reason is for the panels at the bottom that let you see the code, assets, and params.

|Before|After|
|-|-|
|A screenshot of the "Examples" section of the ActionList component Rails docs, showing an embedded version of the playground preview. At the top are two select boxes, one for selecting the preview and the other for choosing the theme (light, dark, etc). There are no additional panels or options.|A screenshot of the "Examples" section of the ActionList component Rails docs. The majority of the screenshot is taken up by an embedded Lookbook preview showing ActionList's playground example rendered in dark mode. Below the preview content are three tabs marked "Source," "Params," and "Assets," just as you would see in Lookbook itself.|


keithamus (approved):
Nice

jonrohan (approved):
Works 🙌🏻 ⚡

PR #610, opened 2023-09-29
Enable linking to page anchors

camertron:
I'm currently trying to set up redirects from the old docsites for our native clients (desktop, mobile, and CLI) to the new docsite at primer.style/design. The old docsites had their various topics broken up into individual pages while the new docsite has combined whole sections into single pages. This means an old URL like https://primer.style/cli/getting-started/principles maps to https://primer.style/design/native/cli/getting-started#principles. Notice the new URL has a URL fragment, i.e. hash, at the end to indicate the browser should scroll to an element with `id="principles"`. This is fine if you're already on the page and click one of the table of contents links, but doesn't work when redirecting from the old docsite to the new one. Web servers like Apache and IIS (the one we're using) strip off the hash during redirect.

This PR fixes the problem by scrolling to the ID specified in the `?anchor=` query param, which web servers do not strip out. It also dynamically modifies the page's URL by converting the query param into a hash, thus fixing up the URL to make it appear as though the redirect was made to a URL with a hash.


src/components/scroll-to-anchor.tsx
```diff
@@ -0,0 +1,45 @@
...
+      window.addEventListener('load', onPageLoad)
+      // remove the event listener when component unmounts
+      return () => window.removeEventListener('load', onPageLoad)
+    }
+  }, [])
+
+  return (
+    <></>
+  )
+}
```
siddharthkp, line 45: Unless I'm missing something, I would assume that this effect would only run after the page has rendered, javascript has arrived and processed. so I don't think we need to add a listener for page load. 

Tested this out and codespaces and it seems to work well:

```suggestion
export default function ScrollToAnchor() {
  useEffect(() => {

    const url = new URL(window.location.href)

    // only do our thing if the URL doesn't have a hash already
    if (url.hash.length === 0) {
      const anchor = url.searchParams.get('anchor')

      if (anchor) {
        // alter the URL by converting the anchor query param into
        // a hash
        url.searchParams.delete('anchor')
        url.hash = `#${anchor}`
        window.history.pushState(null, "", url)
        document.querySelector(`#${anchor}`)?.scrollIntoView()
      }
    }
  }, [])

  return null
}
```

siddharthkp (approved):
Left an optional optimisation that removes the 500ms delay, if you'd like.

Approving in advance :)

src/components/scroll-to-anchor.tsx
camertron: Cool thanks for this! I tried hard to remove that 500ms delay, but the problem is the `scrollIntoView()` call happens before the browser has laid out the page, i.e. before images have loaded. When they do eventually load, the scroll position changes and the anchor is sometimes pushed below the fold.

I tried your suggestion and it seems to work about as well as any non-delay technique I was able to come up with, so let's go with it.

PR #613, opened 2023-10-04
fix: point to the web archive version of article

francisfuzz:
## Summary

Closes https://github.com/primer/design/issues/614

This updates the reference to `An Alphabet of Accessibility Issues` to the Internet Archive's last available, known, and rendered version, since the original website is no longer reachable.


francisfuzz:
Moved this to draft -- turns out this reference is used in more places than one. Please hold!

francisfuzz:
Ready again 🙇

broccolinisoup (approved):
Thanks so much for this PR 🙏🏻

PR #627, opened 2023-10-12
Adds guidelines about inactive buttons

mperrotti:
Relates to https://github.com/github/primer/issues/2676

Screenshot 2023-11-13 at 1 38 05 PM


pksjce (approved):
Looks good to me!

Sidenote - I found it a little hard to imagine more usecases for the `InactiveButton`. Would it make sense to update some more? The banner example was really good. In case of service outage its a great option to have `InactiveButton`

mperrotti:
There have been changes to the design based on feedback from a11y and other designers. I'll update this PR once I feel confident that everybody is aligned on the direction.

PR #640, opened 2023-10-19
Create link-checker.yml

langermank:
Add broken link checker

Closes https://github.com/github/primer/issues/2493

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -10,6 +10,7 @@ jobs:
         id: link-report
         uses: celinekurpershoek/link-checker@v1.0.2
         with:
-          url: "https://primer.style"
+          url: "https://primer.style/"
+          recursiveLinks: true
       - name: Get the result
         run: echo "${{steps.link-report.outputs.result}}"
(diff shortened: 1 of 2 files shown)


camertron:
Wow, this found a _lot_ of broken links! A number of them look like false-positives tho, like all the twitter.com ones? Is it possible to ignore certain domains?

lesliecdubs:
👋 Should we try to revive this PR and get it merged? Seems like a value-add to have a broken link checker for primer.style, especially as we continue to add and adjust redirects.

> A number of them look like false-positives tho, like all the twitter.com ones? Is it possible to ignore certain domains?

It looks like there is an option for ignoring certain URL patterns, could it help?: https://github.com/celinekurpershoek/link-checker?tab=readme-ov-file#ignorepatterns

jonrohan:
I think this library wasn't what we were expecting. It's a web crawler that runs on github actions. When we tested it, it took more than 40 minutes to run and only would work on the production deployment of the website. We should close this or and rethink our problem with a different approach.

lesliecdubs:
Got it, thanks for explaining why this approach wasn't viable; 40mins to crawl prod deployments doesn't seem ideal. I'll go ahead and close the PR for now but leave the associated issue open for consideration.

PR #647, opened 2023-10-25
Update guidance on icon colors for the ActionMenu items

cmwinters:
There's some ambiguity about whether or not we should recommend different colors for leading icons in the ActionMenu.   

This PR updates the guidance to recommend against using colored icons for anything other than destructive actions. 

CleanShot 2023-10-24 at 16 51 30@2x

### DIFF
--- a/content/components/action-menu.mdx
+++ b/content/components/action-menu.mdx
@@ -253,6 +253,27 @@ Especially when providing single or multi select items that rely on a crossmark 
   </Dont>
 </DoDontContainer>
 
+In order to maintain a consistent apperance, avoid using colored icons as a leading visual. Colored icons should be reserved for destructive actions.   
+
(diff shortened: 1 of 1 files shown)


maximedegreve:
Thank you for adding this @cmwinters. 

This looks good but I have a few improvements we could make.
- Could you make the dropdown less wide since the menu should never go this wide for the content it currently holds. [example](https://primer.style/components/action-menu#open).
- Could we find a more realistic example? Actions currently don't have the ability to be forked, sponsored or starred. If actions was a generic name and not the product you're referring to then a 3 dot or cog icon button would make more sense.
- We also tend to group edit actions. Therefore adding a separator about the delete action would make sense here. ([example](https://primer.style/components/action-menu#alignment))

mperrotti (approved):

PR #649, opened 2023-10-26
Updates styles in asset for inactive menu items

mperrotti:


mperrotti:
I think this is a small enough change to YOLO merge...

PR #650, opened 2023-10-27
Add docs for the ClipboardCopyButton

camertron:
The `ClipboardCopyButton` component was [recently merged](https://github.com/primer/view_components/pull/2335) into PVC. Since there is no React analog, and because it's confusing for there to be `ClipboardCopy` and `ClipboardCopyButton` components, I have chosen to not expose it in the nav. Instead, I'm going to submit a PR to link it from `ClipboardCopy`.


camertron:
The deploy probably isn't going to succeed until PVC v0.14.0 is published, since `ClipboardCopyButton` and its docs haven't been released yet.

PR #656, opened 2023-11-08
Remove status from figma section

lukasoppermann:
## Change
This PR removes the status label from the figma docs page

Before | After
--- | ---
Screenshot 2023-11-08 at 11 22 31 | Screenshot 2023-11-08 at 11 29 32

## Reasoning
We don't really have a lifecycle for Figma components. Once they are added to Figma they are `stable` in regards to Figma. They may not be stable in other implementations, but this has no direct impact on Figma. 

For this reason, having a `stable` label may be confusing for users and does not really bring any benefit.


maximedegreve (approved):
Makes sense!

PR #681, opened 2023-11-27
Use Node v18.18.2 LTS (Iron)

mxdvl:
### What?

Bump Node to v18.18.2 LTS Iron, which has [maintenance support until the end of April 2025](https://github.com/nodejs/release#release-schedule).

### Why?

[Node 16 has reached end-of-life in September 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol/)

PR #682, opened 2023-11-27
Refactor status order

mxdvl:
### What?

Refactor `latestStatusFrom` function, so it leverages the existing `compareStatuses` function.

### Why?

Code reuse and simplification: there is now one source of truth for comparing statuses. Follow-up on #679 

Tested by running the following manual tests:

```js
console.assert(latestStatusFrom(['alpha', 'stable']) === 'stable')
console.assert(latestStatusFrom(['stable', 'alpha']) === 'stable')
console.assert(latestStatusFrom(['nope', 'invalid']) === null)
```

Note that v18 of Node and later [come bundled with a test runner](https://nodejs.org/api/test.html). See #681 for upgrade.

### DIFF
--- a/src/rails-status.js
+++ b/src/rails-status.js
@@ -14,19 +14,10 @@ const statusOrder = [
  * @returns {string | null} latest status if valid, `null` otherwise
  */
 export const latestStatusFrom = (statuses) => {
-  let highestRank = -1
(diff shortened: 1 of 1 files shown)


camertron (approved):
Nice, thank you!

PR #683, opened 2023-11-28
Update SelectPanel tabs image

maximedegreve:
@dipree noticed that one of the images still included an older version of the tabs.

Changed image


dipree (approved):
👍

mperrotti (approved):

PR #688, opened 2023-12-05
Add `<details-dialog>` migration guide

TylerJDev:
This PR adds a migration guide for `<details-dialog>` and encourages using `Primer::Alpha::Dialog` instead. 

This guide mainly stems from the work done within https://github.com/github/accessibility-champions/issues/740, and the [migration guide made for that bash](https://github.com/github/accessibility-champions/blob/b5516f06bb36091834c582d4818ab216b7ea5bf1/docs/dialog-linting-conversion-docs.md).

* Added a new migration guide for `<details-dialog>`. This guide includes a migration path for `<details-dialog>`, which encourages usage of `Primer::Alpha::Dialog`, and aims to explain the differences between the two.

### DIFF
--- a/content/guides/development/rails/migration-guides/primer-dialog.mdx
+++ b/content/guides/development/rails/migration-guides/primer-dialog.mdx
@@ -1,14 +1,18 @@
-This guide will show you how to upgrade from the now deprecated `<details-dialog>` element to the latest [`Primer::Alpha::Dialog`](/components/dialog)
+---
+title: Moving Away From <details-dialog>
+---
+
+This guide will show you how to upgrade from the now deprecated `<details-dialog>` element to the latest [`Primer::Alpha::Dialog`](/components/dialog/rails/alpha/)
 component.
 
-### Moving away from `<details-dialog>`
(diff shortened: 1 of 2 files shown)


content/guides/development/rails/migration-guides/primer-dialog.mdx
```diff
@@ -0,0 +1,161 @@
...
+end
+```
+
+Both `Primer::Alpha::Dialog` and `<details-dialog>` will render similar dialogs, albeit with a few key differences:
+
+* Styling
+* Functionality and semantics
+
+### Styling differences
+`<details-dialog>` and `Primer::Alpha::Dialog` will have some minor style differences, such as with padding and margins that are applied by default.
```
langermank, line 44:
```suggestion
`<details-dialog>` and `Primer::Alpha::Dialog` will have some minor style differences, such as with padding and margins that are applied by default. Please utilize these updated styles and refrain from overriding styles to make it appear like the previous design.
```

Just thinking we might want to make it clear that this is the new design and to please use it 😄 feel free to reword that

langermank (approved):
This is awesome 🤩 just left one small comment

PR #692, opened 2023-12-12
Remove `Filter input` as this component is now part of our internal design system

maximedegreve:
Fixes: https://github.com/github/primer/issues/2912


content/components/data-table.mdx
```diff
@@ -544,7 +544,7 @@ If one of the actions is heavily used, pull it out for easier access. Do not pul
 
 ### Filtering
 
-A [filter input](/components/filter-input) is used to write a query and only show rows that
+A [filter input (internal)](https://ui.githubapp.com/storybook/?path=/docs/recipes-filter--readme) is used to write a query and only show rows that
```
lesliecdubs, line 547:
```suggestion
A [filter input (GitHub staff only)](https://ui.githubapp.com/storybook/?path=/docs/recipes-filter--readme) is used to write a query and only show rows that
```

I believe this is how we typically denote content in primer.style that is inaccessible to non-Hubbers.

lesliecdubs (approved):
I'm not an expert on how content from this repo appears in primer.style so you may want an additional approval, but this generally LGTM

PR #700, opened 2023-12-19
Add guidance for writing minimum-viable component docs

mperrotti:


content/guides/contribute/documentation.mdx
```diff
@@ -169,6 +169,57 @@ The description appears directly under the title of the component. It may be aut
...
+  - [Rails Lookbook](https://primer.style/view-components/lookbook)
+  - [Interface guidelines component documentation](/components)
+  - [Interface guidelines UI pattern documentation](/ui-patterns)
+  - [Interface guidelines Accessibility documentation](/guides/accessibility)
+  - Related components that may have guidelines applicable to your component
+  - Take a look at an existing component that was documented using this approach, such as [Action menu](/components/action-menu)
+
+#### Recommendations for tackling this work quickly and efficiently
+
+- Work in pairs or with a small group. Dedicate 1-2 components per person, with each person in the group is responsible for reviewing the others' PRs.
```
joshblack, line 219:
```suggestion
- Work in pairs or with a small group. Dedicate 1-2 components per person, with each person in the group responsible for reviewing the others' PRs.
```

content/guides/contribute/documentation.mdx
```diff
@@ -169,6 +169,57 @@ The description appears directly under the title of the component. It may be aut
...
+- Timebox each component to x hour(s) for writing, x hour(s) for reviewing, and x hour(s) for writing revisions.
+- If you get stuck or a component clearly needs more thorough documentation in order to be accurate, swap it out for another component and add it to the backlog of components that need more thorough documentation.
```
joshblack, line 221: Wanted to ask what does this part of the bullet point mean? (emphasis mine)

> a component clearly needs more thorough documentation **in order to be accurate**

joshblack (approved):
Looks great! Just had a question 👀

content/guides/contribute/documentation.mdx
mperrotti: I just copy-pasted this from existing documentation written by Emily, but I think I understand. I think it's trying to say "if you can't find the information you need to write good docs, keep it in the backlog until we can get that info and move on to another component".

How does this rewording sound?

```suggestion
- If you feel blocked by a lack of information about the component to write useful documentation, move on to documenting a different component until we can get the context we need.
```

dipree:
Came across this by accident because I have been recently affected by exactly these pages:
- https://github.com/primer/design/blob/main/CONTRIBUTING.md#design-guidelines-template
- https://primer.style/guides/contribute/documentation

The one thing I was noticing is that both aren't really in sync with each other but also not with existing docs pages. I appreciate the clean up here.

Most pages have a structure similar to this:
- Overview
- Anatomy
- Options
- Behaviors (also called [Interactions](https://primer.style/guides/contribute/documentation#interactions))
- Usage guidelines (we don't have that in the guidelines but it exists partially in the content)
- Accessibility

It would be great to have the table that explains the intention of each very briefly, which ones are mandatory and which optional. What I was looking for when searching for this page is a very concise documentation of how the ideal component page looks like, in structure, section content and language. That, I would consider what someone is looking for when contributing or updating component pages.

I'm not a huge fan of the term "minimum-viable" I would love not to see pages only have the bare minimum, that sets the bar far too low. Unless it's an experiment/draft but then we should explicitly point out what is missing.

mperrotti:
@dipree - this PR just moves over existing documentation. Could you create an issue for us to address your feedback in a separate PR?

dipree:
Yes sure @mperrotti

PR #709, opened 2024-01-11
Spinner interface guidelines

mperrotti:
Adds interface guidelines for the Spinner component.

### DIFF
--- a/content/components/spinner.mdx
+++ b/content/components/spinner.mdx
@@ -1,6 +1,6 @@
 ---
 title: Spinner
-description: Use Spinner to let users know that content is being loaded.
+description: Use Spinner to show users that an action is in progress.
 railsIds:
 - Primer::Beta::Spinner
 reactId: spinner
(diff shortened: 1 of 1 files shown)


content/components/spinner.mdx
```diff
@@ -12,11 +12,52 @@ import {Link as GatsbyLink} from 'gatsby'
...
+
+</div>
+
+<img src="https://github.com/primer/design/assets/2313998/b31c4058-929f-471c-860e-6a8c2c6af21e" width="960" role="presentation" />
+
+</Box>
+
+## Accessibility
+
+By default, our spinner component provides a generic loading status announcement for assistive technologies. We recommend that you specify a custom announcement specific to your use-case. For example: "Loading changed files".
```
keithamus, line 51: It this an aspirational design doc or does this work today? If it's aspirational, have we checked with a11y that this is the right thing to do? My spidey senses are tingling on this prose.

keithamus (approved):

PR #710, opened 2024-01-17
Migrate docs for deprecated React components

camertron:
## Deprecated component status

This PR migrates docs for deprecated React components from the old primer.style/react site to the new docsite. This was not a straightforward migration, since the current docs exist within this matrix of sadness:

|Component|Has non-dep counterpart|Bespoke .md file|components.json?|Type of props table|
|-|-|-|-|-|
|`ActionList`|✅|✅|🚫|Hand-written md
|`ActionMenu`|✅|✅|🚫|Hand-written md|
|`Button`|✅|✅|🚫|Hand-written React|
|`FilterList`|🚫|✅|✅|Imports .data.json|
|`FilteredSearch`|🚫|✅|✅|Imports .data.json|
|`SideNav`|🚫|✅|🚫|Hand-written md|
|`UnderlineNav`|✅|✅|🚫|Imports .data.json|

Legend:

* _Has non-dep counterpart_: The deprecated component has a newer, non-deprecated version.
* _Bespoke .md file_: At least some of the docs for the deprecated version of the component live in a separate, individually-maintained .md file that contains additional information to what may be available in components.json.
* _components.json?_: The deprecated component is listed in components.json, with a status of `Deprecated`.
* _Type of props table_: How the props table is written in the bespoke .md file. "Hand-written React" means it uses `<PropsTable>` component. "Imports .data.json" means it imports the corresponding component_name.data.json file and passes the data to the `<ComponentProps>` component.

## Action taken

1. React components can now have multiple versions. Each version can be reached by way of a dropdown menu that appears on each component page. The URLs for these components take the form /components/component-name>/react/status, where the status is eg. "alpha," "beta," "draft", "deprecated," etc. This follows the same URL convention as Rails components. A special URL, /components/component-name/react/latest, redirects to the most stable version of the component, usually "alpha."
2. For `ActionList`, `ActionMenu`, `Button`, and `UnderlineNav`, I copy/pasted the existing docs into dedicated pages under components/component-name/react/deprecated, which follows the same naming conventions I described above. I added a special `reactStatus: deprecated` field to the page's YAML frontmatter to mark components that have dedicated pages for their deprecated versions, which gatsby-node.js queries when generating pages.
3. Since `FilterList` and `FilteredSearch` are listed in components.json, I elected to turn them into top-level /deprecated-components, i.e. under the "Deprecated components" section in the nav. Although there is some additional information in their bespoke .md files, it's minimal enough to drop from the new docsite.
4. `SideNav` now lives under the top-level "Deprecated components" nav section. There was enough information in its bespoke .md file that I decided to port it over by creating /deprecated-components/side-nav/react/latest.md, which is consistent with our now site-wide URL structure.

### Prop tables

1. For hand-written markdown, I copied the props table over without modification.
2. For hand-written React, I copied the props table over without modification. I also copied over the `<PropsTable>` component from Primer React.
3. The contents of the .data.json files used by `FilterList` `FilteredSearch` are also part of components.json. By opting to use react-component-layout.tsx for these components, the props table is now generated automatically. The results should appear identical.

## Additional hurdles

Because the new docsite uses an outdated version of Gatsby, we are unable to use modern versions of Primer React to build it. As a direct consequence, I had to re-implement the old docsite's `ThemeReferenceTree` component, which relies on Primer React's `TreeView` component. Sadly `TreeView` isn't available in older versions of the library. I had to rewrite it to use `<summary>`/`<details>` elements instead.


keithamus (approved):
:shipit:

PR #317, opened 2022-10-25
[ActionBar] Support mixed items

simurai:
This is a follow-up to https://github.com/primer/design/pull/287. It adds support for mixed items.

Before | After
--- | ---
Existing Action bar | Updated Action bar that also supports Button components
`IconButton` only | `IconButton` + `Button`
[Current docs](https://primer.style/design/components/action-bar) | 👀 [Preview](https://primer-3a209d8009-26441320.drafts.github.io/components/action-bar)

Part of https://github.com/github/primer/issues/1438

PR #726, opened 2024-02-21
Update action bar docs

mperrotti:
We're beginning to port the ActionBar component from Primer ViewComponents to Primer React Components. This will help us provide a sturdy and accessible ActionBar to be used in the comment box and the ListView component.

Relevant links:
- https://github.com/github/primer/issues/3020
- [ActionBar in Primer ViewComponents Lookbook](https://view-components-storybook.eastus.cloudapp.azure.com/view-components/lookbook/inspect/primer/alpha/action_bar/default)
- [ActionBar in Primer Recipes Storybook](https://ui.githubapp.com/storybook/?path=/story/recipes-actionbar--action-bar-toolbar-variant)


mperrotti:
As I was updating our guidelines, I came up with the following questions:
- Was there a strong reason to restrict ActionBar to only use icon buttons?
  - If so, how might we support toolbars like the one in ListView ([screenshot](https://github.com/primer/design/assets/2313998/612e8e60-98b9-4040-b203-41623de6302f) from [ActionBar epic](https://github.com/github/primer/issues/3020))
- Should we restrict ActionBar buttons to only use `invisible` and `secondary` variants?
  - If so, do other folks agree that this be configured as a prop on the ActionBar?
  - If so, are we comfortable with the fact that we can't use `danger` for destructive actions in the ActionBar?
- Will adding space between each button cause problems with our current (more compact) usage of ActionBar?
  - If we support the `secondary` variant, we'll need to add space between the buttons so we don't have two bordered buttons bumping into each other.
  - Similarly, `invisible` buttons with text labels will need more space between them so that the button labels don't run together.
- The Primer Recipes ActionBar has a "menu" variant ([Storybook](https://ui.githubapp.com/storybook/?path=/story/recipes-actionbar--action-bar-menu-variant&args=variant:menu)). When should consumers use this instead of the "toolbar" variant?

---

I'm keeping this PR as a "Draft" until we can get aligned on these and can update the docs accordingly.

simurai (commented):
> * Was there a strong reason to restrict ActionBar to only use icon buttons?

Not really, there was even a PR https://github.com/primer/design/pull/317 to add other button types. But at the time we only needed icon buttons for the CommentBox, https://github.com/github/primer/issues/1438#issuecomment-1437736932.

> * Should we restrict ActionBar buttons to only use `invisible` and `secondary` variants?

I think it's ok to allow other variants as long as it can be mapped to an ActionMenu item (will be used when not enough space). So `danger` should be fine too.

> * Will adding space between each button cause problems with our current (more compact) usage of ActionBar?

Yeah, this will be a problem.

- Maybe solvable by using a sibling selector `.item + .item {add margin-left}`.
- Or by introducing "groups" that have a divider line. And with the restriction that inside each group, all buttons need the same variant. Then we could use `gap` for spacing.

> * The Primer Recipes ActionBar has a "menu" variant ([Storybook](https://ui.githubapp.com/storybook/?path=/story/recipes-actionbar--action-bar-menu-variant&args=variant:menu)). When should consumers use this instead of the "toolbar" variant?

Haven't seen this variant. It looks like "Toolbar" can use any kind of child component. But I assume it won't work with the overflow menu. Like the segmented control might be hard to place inside a ActionMenu. So it's more meant for use when space is guaranteed to be enough.

---

I think in general, it's easier to add something when really needed than to remove something later that turns out is not used much or causes issues. So trying to keep components simple as long as we can.

mperrotti:
> - Maybe solvable by using a sibling selector `.item + .item {add margin-left}`.
> - Or by introducing "groups" that have a divider line. And with the restriction that inside each group, all buttons need the same variant. Then we could use `gap` for spacing.

@simurai - I don't understand how either of these solutions will help us keep the ActionBar more compact. Could you elaborate?

simurai:
> @mperrotti I don't understand how either of these solutions will help us keep the ActionBar more compact. Could you elaborate?

Ah.. no. Adding extra space would make it **less** compact. I think I misunderstood that "not keeping it compact" is the "problem".

Maybe that's something the user of the component would have to decide:

- If the available space is tight (e.g. comment box), only use icons.
- If there is plenty of available space, then consider using `secondary` buttons with text labels.

In the [closed proposal](https://github.com/primer/design/pull/317), there was even a [spacious variant](https://primer-3a209d8009-26441320.drafts.github.io/components/action-bar#gap) to space `secondary/danger` buttons out even more:

Screenshot 2024-02-22 at 11 37 45

content/components/action-bar.mdx
```diff
@@ -74,7 +105,16 @@ Dividers can be added to visually group related buttons.
 
 ### Overflow menu
 
-When the buttons don't fit in the available space, an overflow button ("kebab" icon) is added at the end of the action bar signaling that there are more actions available. Clicking on the overflow button opens a menu with the remaining actions that didn't fit.
+{/* <!-- TODO: determine how/if we want to support hiding items out of order --> */}
```
mperrotti, line 108: I'm wondering if we want to support the ability to _always_ hide specific items. This was explored in an earlier design iteration that was abandoned.

Should we start to support this now? Or keep the behavior as-is?

If we do choose to support this, how should we let users specify actions that are always hidden?

1. Rely on the order the buttons are passed in, and let users specify a maximum number of action bar buttons
2. Let consumers specify which actions should go in the overflow menu regardless of what order they're passed in

mperrotti: I spent some time exploring this functionality. Please excuse (or enjoy) the lo-fi ASCII art mocks 😅 

## Truncate after max number of items

Assume the max number of items was set to `5`.

### Option A

Limit how many buttons can be shown and rely on order. Any buttons that come after the limit are placed in the overflow menu.

#### Wide screen

```
[a] [b] | [c] [d] | [e] [...]
    |-----------------------|
    | [f]                   |
    | [g]                   |
    | [h]                   |
    | [i]                   |
    |-----------------------|
```

#### Narrow screen

```
[a] [b] | [c] [...]
    |-------------|
    | [d]         |
    | ------------|
    | [e]         |
    | [f]         |
    | [g]         |
    | [h]         |
    | [i]         |
    |-------------|
```

## Allow items to be explicitly overflowed in any order

Assume the buttons were passed in alphabetical order.

Assume that anything that appears in the overflow menu in a "wide screen" has been explicitly set to only appear in the overflow menu.

### Option A - use divider position to prepend to section:

#### Wide screen

```
[a] [b] | [e] [g] [i] [...]
    |---------------------|
    | [c]                 |
    | [d]                 |
    |---------------------|
    | [f]                 |
    | [h]                 |
    |---------------------|
```

#### Narrow screen

```
[a] [b] | [e] [...]
    |-------------|
    | [c]         |
    | [d]         |
    | ------------|
    | [g]         | <- prepended "g"
    | [i]         | <- prepended "i"
    | [f]         |
    | [h]         |
    |-------------|
```

### Option B (use divider position and order of button child for order of menu item):

#### Wide screen

```
[a] [b] | [e] [g] [i] [...]
    |---------------------|
    | [c]                 |
    | [d]                 |
    |---------------------|
    | [f]                 |
    | [h]                 |
    |---------------------|
```

#### Narrow screen

```
[a] [b] | [e] [...]
    |-------------|
    | [c]         |
    | [d]         |
    | ------------|
    | [f]         |
    | [g]         | <- added "g" to section based on the order it was passed
    | [h]         |
    | [i]         | <- added "i" to section based on the order it was passed
    |-------------|
```

content/components/action-bar.mdx
```diff
@@ -74,7 +105,16 @@ Dividers can be added to visually group related buttons.
...
+
+When actions are put in the overflow menu, an icon button (["kebab" icon](/foundations/icons/kebab-horizontal-16)) is added at the end of the action bar signaling that there are more actions available. Clicking on the overflow button opens a menu with the remaining actions that didn't fit.
+
+Dividers from the action bar are preserved in the overflow menu.
+
+There are two ways that actions can be pushed into the overflow menu:
+
+- When the buttons don't fit in the available horizontal space
+- By explicitly defining actions that will only appear in the overflow menu
```
mperrotti, line 117: Delete this if we choose not to let consumers force actions to only appear in the overflow menu.

content/components/action-bar.mdx
```diff
@@ -223,10 +278,16 @@ An action bar has an ARIA role of [`toolbar`](https://developer.mozilla.org/en-U
...
 | `Enter` or `Space` | Triggers the button **action**.                                                                                                                                                                                                                                                                                                                        |
 
+The `toolbar` role should be overridden if you're using an action bar in a context where the behavior of left arrow, right arrow, "Home", and "End" keys conflict with their behavior in a `toolbar`. For example, as list item actions within a list where focus may be moved between items using arrow, "Home", and "End" keys.
+
+In cases where we don't use the `toolbar` role, no role is applied and each buttons has one tab stop (their default keyboard navigation behavior).
+
 <!-- TODO: Add video -->
 
 ### Touch targets
```
mperrotti, line 287: The PVC implementation doesn't follow this guidance. An issue has been created: https://github.com/github/primer/issues/3078

content/components/action-bar.mdx
pksjce: Are there usecases where certain items always show up in the overflow menu. I would say the comment box actionbar mostly does "Truncate after max number of items" and that too based on available space. If all items can be shown on the action bar then we simply show them. As the container reduces in space, we pop them on to the overflow menu.

content/components/action-bar.mdx
pksjce: Yeah, lets implement this only if we find a usecase.

content/components/action-bar.mdx
```diff
@@ -168,6 +208,13 @@ Action bars can have 3 different sizes:
 - Medium (`32px`) (default)
 - Large (`40px`)
 
+### Button variants
+
+Icon buttons should default to the `invisible` (no border/background) variant. Buttons with text labels should default to the `secondary` variant.
+
+You may use the `danger` variant for destructive actions. The `primary` variant should not be used in an action bar because primary page actions are highly unlikely to appear in a collection with other actions.
+
```
pksjce, line 216: The API could be
- ActionBar.IconButton (default to invisible, cannot accept text)
- ActionBar.TextButton (default to secondary, can accept icon?)
- ActionBar.DangerButton (default to danger, can accept icon?)

content/components/action-bar.mdx
```diff
@@ -223,10 +278,16 @@ An action bar has an ARIA role of [`toolbar`](https://developer.mozilla.org/en-U
...
+The `toolbar` role should be overridden if you're using an action bar in a context where the behavior of left arrow, right arrow, "Home", and "End" keys conflict with their behavior in a `toolbar`. For example, as list item actions within a list where focus may be moved between items using arrow, "Home", and "End" keys.
+
+In cases where we don't use the `toolbar` role, no role is applied and each buttons has one tab stop (their default keyboard navigation behavior).
+
 <!-- TODO: Add video -->
 
 ### Touch targets
 
+When the user is on a device with coarse pointer (such as a touch screen), the button target needs to be enlarged so users can easily tap the correct target button.
+
```
pksjce, line 290: How does the touchtarget work with different recommended `ActionBar` sizes?

mperrotti: I described a possible solution in the description of this issue: https://github.com/github/primer/issues/3078

I want to get opinions from other designers before committing to that direction.

content/components/action-bar.mdx
mperrotti: I think we might be able to have a single component to handle all three. An IconButton will need text _somewhere_ for screen readers and tooltips.

mperrotti: If we want to be consistent with the Button API, we could keep `ActionBar.IconButton` separate

pksjce: `IconButton` does take `aria-label` which I'm currently using for the text when I pop it into the menu on overflow.
But I'm just torn between allowing variants broadly on `ActionBar.Button` or make different components. Let me think more on it and consult.

siddharthkp: > If we want to be consistent with the Button API, we could keep ActionBar.IconButton separate

+1 to that. I personally like the idea of keeping the names consistent with the Button API

1) `<ActionBar.IconButton icon={PaperclipIcon} aria-label="Attach text" />`
2) `<ActionBar.Button leadingVisual={LabelIcon}>Set Label</ActionBar.Button>`
3) `<ActionBar.Button variant="danger" leadingVisual={LabelIcon}>Set Label</ActionBar.Button>`

We could, of course, narrow down the props in ActionBar.Buttons to only support a subset of the Button props. For example, 
1) a limited set of variants instead of the whole set
2) only supporting leadingVisual and not trailingVisual or count. (bonus: a shortcut prop called `icon` instead of `leadingVisual`)
3) allow `size` on ActionBar, but not on ActionBar.Buttons

content/components/action-bar.mdx
siddharthkp: This is really cool! Do we already have examples from production or recipes where we do this?

> I'm wondering if we want to support the ability to always hide specific items.

That's interesting! I don't think I've come across this yet. (Or maybe, just never noticed it)

siddharthkp: Another question here, I assume for 99% of the use cases, we would want to follow the same order in wide and narrow (like in comment box and ListView). But, there could be scenarios where that's not true. For example, always keeping the destructive action in the end, but keeping it visible.

icon buttons in action bar, destructive button is always visible as the last item

I think we can could come up with a custom prop that keeps the 99% case still pretty, something like `ActionBar.Icon icon={TrashIcon} avoidOverflow={true}`

siddharthkp: For the use case / customisation where the order outside the overflow isn't the same as the order inside the overflow: That's interesting! I don't think I've come across this yet. (Or maybe, just never noticed it)

I didn't fully understand the examples with divider here. 🤔 Are we inferring information based on the divider or is there explicit information about order passed here.

content/components/action-bar.mdx
pksjce: Yes! I am fine with this reco thanks!

pksjce (approved):

content/components/action-bar.mdx
mperrotti: > We could support something like <ActionBar.IconButton icon={TrashIcon} avoidOverflow={true} /> which would keep this icon visible until there are multiple avoidOverflow competing for space and then we start overflowing the last one.

> I think we should only pick one of these for the API, either avoidOverflow or preferOverflow, not both at the same time!

I like this functionality - it makes a lot of sense. I'm not sure if this is in scope though. What do you think @pksjce ?

I think we should go with `avoidOverflow` since the default value would be `false`.

---

> I didn't fully understand the examples with divider here. 🤔 Are we inferring information based on the divider or is there explicit order passed to ActionBar?

The only information we're inferring based on the ActionBar divider is that there should be a divider in the same spot in the ActionMenu.

simurai: > @siddharthkp We could support something like `<ActionBar.IconButton icon={TrashIcon} avoidOverflow={true} />` which would keep this icon visible until there are multiple `avoidOverflow` competing for space and then we start overflowing the last one.
> @mperrotti I think we should go with `avoidOverflow` since the default value would be `false`.

I like this option too. ❤️ The example with the global nav kinda needs this behavior if it wants to use `ActionBar`.

PR #733, opened 2024-02-27
Adds guidance about how to use aria-disabled with inactive controls

mperrotti:
I've added docs that explain why we convey "disabled" and "inactive" controls differently to assistive technologies.

There has been some confusion about why we don't default to using `aria-disabled` with inactive controls such as buttons and action menu items.

---

I'm keeping these changes in a "draft" state until we determine whether we want to keep the implementation as-is, or update it. Related issue: https://github.com/github/primer/issues/3045

### DIFF
--- a/content/components/action-menu.mdx
+++ b/content/components/action-menu.mdx
@@ -466,6 +466,22 @@ See [focus management](/guides/accessibility/focus-management) for more informat
   </Box>
 </Box>
 
+### Inactive menu items vs disabled menu items
+
+Since inactive items can still be focused and activated (for example, to show a modal with more information about an outage), they should not be conveyed as disabled using `aria-disabled`.
+
+This behavior is consistent with the behavior of inactive buttons. See the [button component accessibility guidelines](/components/button#inactive-buttons-and-aria-disabled) and [inactive button guidelines](/components/button#inactive) for more details.
+
+#### Inactive menu items:
+
+- Can be activated
+- Need to be able to get focus so screen reader users can "read" the message about it being inactive
+
+#### Disabled menu items:
+
+- Do not get focus
+- Cannot be activated
+
 ### Keyboard navigation
 
 #### Trigger button
--- a/content/components/action-menu.mdx
+++ b/content/components/action-menu.mdx
@@ -468,7 +468,7 @@ See [focus management](/guides/accessibility/focus-management) for more informat
 
 ### Inactive menu items vs disabled menu items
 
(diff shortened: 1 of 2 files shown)


content/components/action-menu.mdx
```diff
@@ -466,6 +466,22 @@ See [focus management](/guides/accessibility/focus-management) for more informat
...
+This behavior is consistent with the behavior of inactive buttons. See the [button component accessibility guidelines](/components/button#inactive-buttons-and-aria-disabled) and [inactive button guidelines](/components/button#inactive) for more details.
+
+#### Inactive menu items:
+
+- Can be activated
+- Need to be able to get focus so screen reader users can "read" the message about it being inactive
+
+#### Disabled menu items:
+
+- Do not get focus
```
TylerJDev, line 482: I'm wondering if this is true in all cases. If a menu item receives `aria-disabled`, wouldn't we want to keep it focusable? Example: 

```html
<ul role="menu">
  <li role="menuitem" tabindex="0" aria-disabled="true">Item 1</li>
  <li role="menuitem" tabindex="-1">Item 2</li>
</ul>
```

Would the markup differ than the example?

content/components/button.mdx
```diff
@@ -436,6 +436,12 @@ There are rare cases where it's ok to disable a button, but it should generally
 
 For more information on disabled controls, see GitHub's [link and button a11y guidance](https://accessibility-playbook.github.com/link-and-button-guidance#disabling-links-and-buttons) (GitHub staff only).
 
+### Inactive buttons and `aria-disabled`
+
+An [inactive](#inactive) button **should not be conveyed as disabled** with `aria-disabled` if it performs an action when activated. For example, showing a dialog with more info about why the button is inactive.
+
+An [inactive](#inactive) button **may be conveyed as disabled** with `aria-disabled` if it does not perform an action when activated.
```
TylerJDev, line 443: Would the component determine if `aria-disabled` should be applied in cases where an inactive state is used without an action, or is that the responsibility of the developer? If it's the latter, _should_ we make this part of the component?

TylerJDev (commented):
Nice! ✨ I just had a few comments, none are blocking!

content/components/action-menu.mdx
mperrotti: We might, but currently we only pass `aria-disabled` for disabled items. I don't believe we allow consumers to manually pass `aria-disabled`.

content/components/button.mdx
mperrotti: It's the latter. We could make it part of the component.

ericwbailey (approved):

PR #737, opened 2024-03-07
Fix typo on the Radio page

adampage:
Just a silly little fix for [a typo on the Radio page](https://primer.style/components/radio#:~:text=be%20used%20on-,it%27s,-own%3A%20it%20must). Thank you for such a well-documented and inspiring design system. 🙏🏻


keithamus (approved):
Great catch, thanks!

PR #741, opened 2024-03-12
Fix small grammer mistake in Update button.mdx

dominik-ilja:


keithamus (approved):
Another nice catch, thanks!

PR #746, opened 2024-03-22
Adding docs for upcoming inlineMessage component

lukasoppermann:
Waiting for https://github.com/primer/react/pull/4443 to merge & publish

PR #748, opened 2024-03-22
Add docs for new banner

lukasoppermann:
Only merge when the react component is implemented

### DIFF
--- a/content/components/banner.mdx
+++ b/content/components/banner.mdx
@@ -15,7 +15,7 @@ import {AccessibilityLink} from '~/src/components/accessibility-link'
 
 <img
   width="960"
-  alt="Image displaying each variant of banenr, from top: critical, warning, success, info, upsell."
+  alt="Image displaying each variant of banner, from top: critical, warning, success, info, upsell."
   src="https://github.com/primer/primitives/assets/813754/f96b1eda-2413-4fc1-b1a7-3775b0c3b324"
 />
 
@@ -32,7 +32,7 @@ For system wide information, consider using an [announcement](../ui-patterns/mes
 <img
   width="960"
   alt="Anatomy of a banner, including the message, title, icon, dismiss button, primary, and secondary action"
-  src="https://github.com/primer/design/assets/813754/defa52f8-49bb-4332-aaa1-872a22a563a9"
+  src="https://github.com/primer/design/assets/813754/40ff27e0-8d69-490a-a619-a4956828e77e"
 />
 
 ## Options
--- a/content/components/banner.mdx
+++ b/content/components/banner.mdx
@@ -16,7 +16,7 @@ import {AccessibilityLink} from '~/src/components/accessibility-link'
 <img
   width="960"
   alt="Image displaying each variant of banner, from top: critical, warning, success, info, upsell."
-  src="https://github.com/primer/primitives/assets/813754/ec62d3e4-a61b-465a-9361-ee7c46d59715"
+  src="https://github.com/primer/design/assets/813754/1f221ea9-0a82-4f4f-a8eb-b1d61ef1903d"
 />
 
 ## Usage
--- a/content/components/banner.mdx
+++ b/content/components/banner.mdx
@@ -151,7 +151,6 @@ Don't show more than one banner at a time. For more information, see this articl
     src="https://github.com/primer/primitives/assets/813754/c5cc31d6-dafc-426d-999b-215f68ca7c28"
   />
   <Box as="p" mt="0">
-    <h4>Title</h4>
     Use a title to consisly present the topic of the banner when the message of the banner is complex and too long to scan.
   </Box>
 </Box>
@@ -171,8 +170,10 @@ Don't show more than one banner at a time. For more information, see this articl
     src="https://github.com/primer/primitives/assets/813754/85c94192-eb1a-4e7c-8fc0-b3c5f31ac251"
   />
   <Box as="p" mt="0">
-    <h4>Dismiss button</h4>
-    Banners with an <i>onDismiss</i> property automatically show a dismiss button.
+  
+  <h4>Dismiss button</h4>
+  Banners with an `onDismiss` property automatically show a dismiss button.
+  
   </Box>
 </Box>
 <Box
@@ -188,11 +189,11 @@ Don't show more than one banner at a time. For more information, see this articl
     src="https://github.com/primer/primitives/assets/813754/a8684eed-5544-4d6d-a1c6-eda1376e95b5"
   />
   <Box as="p" mt="0">
-    <h4>Critical banner</h4>
+    <h4>Exception: Critical banner</h4>
     <p>
-    Critical banner can't be made dismissable. Since they should only be used for errors that need to be resolved, the only way to get rid of a critical banner is to resolve the error. If they are used to warn the user about destructive actions the banner should also not be dismissable.</p>
+    Critical banners can't be dismissable. They are mostly used for errors which need to be resolved. As a result you can only get rid of a critical banner by resolving the error. If used to warn about destructive actions, the banner should also not be dismissable.</p>
     <p>
-    If the error must be resolved on a different page, it is recommended to add an action that directs the user to the page where the error can be resolved.
+    If the error must be resolved on a different page, use an action that directs the user to the page where the error can be resolved.
     </p>
   </Box>
 </Box>
@@ -212,7 +213,6 @@ Don't show more than one banner at a time. For more information, see this articl
     src="https://github.com/primer/primitives/assets/813754/7c57bc61-ee0d-4673-86e9-550b945b117a"
   />
   <Box as="p" mt="0">
-    <h4>Actions</h4>
     <p>A maximum of 2 actions can be added to a banner. Whenever possible it is recommended to only add a single action to better direct the users attention.</p>
     <p>For banners with a single action, you can choose between adding a primary or secondary action depending on how much attention you want to draw to the action.</p>
     <p>For banners with two actions, one of them must be primary and the other must be secondary. In this case, use the primary action for the recommended action. For example, the primary action enables a feature and the secondary action links more information.</p>
@@ -234,8 +234,9 @@ Don't show more than one banner at a time. For more information, see this articl
     src="https://github.com/primer/primitives/assets/813754/9de199af-6488-4ae3-b4a7-14896f5cb719"
   />
   <Box as="p" mt="0">
-    <h4>Icon</h4>
-    The icon relates to the selected <a href="#variant">variant</a> and can't be disabled. It can only be changed for banners of variant <i>info</i> or <i>upsell</i>.
+  
+  The icon relates to the selected <a href="#variant">variant</a> and can't be disabled. It can only be changed for banners of variant `info` or `upsell`.
+  
   </Box>
 </Box>
 
--- a/content/components/banner.mdx
+++ b/content/components/banner.mdx
@@ -172,6 +172,7 @@ Don't show more than one banner at a time. For more information, see this articl
   <Box as="p" mt="0">
   
   <h4>Dismiss button</h4>
+
   Banners with an `onDismiss` property automatically show a dismiss button.
   
   </Box>
--- a/content/components/banner.mdx
+++ b/content/components/banner.mdx
@@ -30,7 +30,7 @@ For system wide information, consider using an [announcement](../ui-patterns/mes
 ### Best practices
 Use banners sparingly and only when necessary. Banners can disrupt the user experience and should only be used when and where relevant.
 
-Keep the message concise and direct. Don't use headings or multiple type sizes in banners, rely on the default paragraph size and use the title when appropriate.
+Keep the message concise and direct. Don't use heading styling or multiple type sizes in banners, rely on the default paragraph size and use the title when appropriate.
 
 Don't show more than one banner at a time. For more information, see this article on [alert fatigue](https://www.nngroup.com/videos/alert-fatigue-user-interfaces/).
 
@@ -55,7 +55,7 @@ Don't show more than one banner at a time. For more information, see this articl
 >
   <img
     width="456"
-    alt="critical banner"
+    alt="Text within a box that has a light red outline and a lighter red background. There is a vibrant red stop icon with an exclamation point inside that comes before the text within the banner."
     src="https://github.com/primer/primitives/assets/813754/9c245a1e-5d59-44bf-8574-aedb06a286b7"
   />
   <div>
@@ -76,7 +76,7 @@ Use critical banners to signal critical errors, system failures, or necessary de
 >
   <img
     width="456"
-    alt="warning banner"
+    alt="Text within a box that has a yellow outline and a lighter yellow background. There is a dark yellow triangle icon with an exclamation point inside that comes before the text within the banner."
     src="https://github.com/primer/primitives/assets/813754/afa299a9-4efe-41c4-a8cd-4e166d93e750"
   />
   <div>
@@ -97,14 +97,14 @@ Use warning banners to alert users about potential issues or impactful changes. 
 >
   <img
     width="456"
-    alt="success banner"
+    alt="Text within a box that has a green outline and a lighter green background. There is a vibrant green circle icon with a check icon inside the circle that comes before the text within the banner."
     src="https://github.com/primer/primitives/assets/813754/5bc8b2b6-f935-4153-809a-00282ec0e1a3"
   />
   <div>
 
 #### Success
 
-Use success banners to indicate task completion or successful actions. For example, to confirm settings have been saved or a successful issue transfer. Only use success messages if the success is not easily apparent in the UI.
+Use success banners to indicate task completion or successful actions. For example, a successful issue transfer when the user is not brought to the new page. Only use success messages if the success is not easily apparent in the UI.
 
   </div>
 </Box>
@@ -118,7 +118,7 @@ Use success banners to indicate task completion or successful actions. For examp
 >
   <img
     width="456"
-    alt="info banner"
+    alt="Text within a box that has a light blue outline and a lighter blue background. There is a vibrant blue circle icon with the letter i inside that comes before the text within the banner."
     src="https://github.com/primer/design/assets/813754/172e3eb4-35ea-4d75-a015-6d00d3ebd914"
   />
   <div>
@@ -139,7 +139,7 @@ Use info banners to inform about non-critical context and information. For examp
 >
   <img
     width="456"
-    alt="upsell banner"
+    alt="Text within a box that has a light purple outline and a lighter purple background. There is a vibrant purple computer icon that comes before the text within the banner."
     src="https://github.com/primer/primitives/assets/813754/fca5f277-6baf-491d-96a8-6f23dd18d49a"
   />
   <div>
@@ -168,7 +168,9 @@ Use upsell banners to inform the users about a feature that can be enabled by up
 
 Use a title to concisely present the topic of the banner when the message of the banner is complex and too long to scan. 
 
-A title is required and will be used as the accessibility title. However, you can visually hide the title. It is recommended to hide the title if the body of the banner is short and easily scannable.
+A title is required and will be used as a semantic page heading. However, you can visually hide the title. It is recommended to hide the title if the body of the banner is short and easily scannable.
+
+Use the built-in accessibility annotations in the Figma components to define the heading level and text relevant to its context. 
 
 </Box>
 
@@ -190,7 +192,7 @@ A title is required and will be used as the accessibility title. However, you ca
 
 #### Dismiss button
 
-Banners with an `onDismiss` property automatically show a dismiss button. This should be used for banners that inform about something that the user can not solve. For example issues with the organization they are part of or informational banners. When it makes sense, the choice to dismiss the banenr should be stored and the banner should not be shown again.
+Banners with an `onDismiss` property automatically show a dismiss button. This should be used for banners that inform about something that the user can not solve. For example issues with the organization they are part of or informational banners. When it makes sense, the choice to dismiss the banner should be stored and the banner should not be shown again.
 
   </div>
 </Box>
@@ -253,7 +255,7 @@ For banners with two actions, one of them must be primary and the other must be 
 >
   <img
     width="456"
-    alt="banners with icons, some of which are customized"
+    alt="Three banners with leading icons. Success banner has its standard circle and check icon, upsell banner has a custom computer icon, and the info banner has a custom shield icon with a keyhole inside of the lock."
     src="https://github.com/primer/primitives/assets/813754/9de199af-6488-4ae3-b4a7-14896f5cb719"
   />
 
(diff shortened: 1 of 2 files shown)


content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+  mb={3}
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column', 'column', 'column', 'column', 'row']}
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="banner without title and banner with title"
+    src="https://github.com/primer/design/assets/813754/c5876803-7b5b-486d-b2ec-27fcfdf0c5d8"
```
mperrotti, line 144: I think this example image would work better if we either:
- Only showed the banner with a title
- Labeled each banner as "With a title" and "Without a title"

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="trailing actions on a banner"
+    src="https://github.com/primer/design/assets/813754/b4d2e88e-2f0e-4b29-8f21-13ac8399c943"
+  />
+  <Box as="p" mt="0">
+    <h4>Actions</h4>
+    A maximum of 2 actions can be added to a banner. Whenever possible it is recommended to only add a single action to better direct the users attention.
```
mperrotti, line 209: The text after the first paragraph is being rendered as a code block. This can be resolved by following my MDX formatting suggestion in the inline message guidelines PR: https://github.com/primer/design/pull/746#discussion_r1536118986

Screenshot 2024-03-22 at 4 05 20 PM

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+
+    You can choose between adding a primary or secondary action, depending on how much attention you want to draw to the action.
```
mperrotti, line 211:
```suggestion
    For banners with a single action, you can choose between adding a primary or secondary action depending on how much attention you want to draw to the action.
```

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+
+    If you do need two actions in your banner, one of them must be primary and the other must be secondary. In this case, use the primary action for the recommended action for example, to enable a feature and the secondary one to get more information.
```
mperrotti, line 213:
```suggestion
    For banners with two actions, one of them must be primary and the other must be secondary. In this case, use the primary action for the recommended action. For example, the primary action enables a feature and the secondary action links more information.
```

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+    alt="banners with icons, some of which are customized"
+    src="https://github.com/primer/design/assets/813754/dd6d9159-8de7-4f71-b69b-ede15284aeb4"
+  />
+  <Box as="p" mt="0">
+    <h4>Icon</h4>
+    The icon relates to the selected <a href="#variant">variant</a> and can't be disabled. It can only be changed for banners of variant <i>info</i> or <i>upsell</i>.
+  </Box>
+</Box>
+
+## Breakpoints
```
mperrotti, line 237: I'd recommend nesting this section under "Layouts" and renaming this header to "Responsive design".

This is just my personal preference and should not be considered blocking feedback.

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+
+On large screens the layout of the elements within the banner adjust depending on which elements are enabled. On small screensthe banner always uses the stacked layout.
+
+## Layouts
+<img
+  width="960"
+  alt="Layouts depending on which elements are enabled"
+  src="https://github.com/primer/design/assets/813754/fe702f23-de88-47a9-b703-b8b262698a50"
+/>
+Text missing
```
mperrotti, line 252:
```suggestion
```

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+## Breakpoints
+<img
+  width="960"
+  alt="Anatomy of a banner, including the message, title, icon, dismiss button, primary, and secondary action"
+  src="https://github.com/primer/design/assets/813754/b1db8982-43ce-4a95-b25a-407020b53440"
+/>
+
+On large screens the layout of the elements within the banner adjust depending on which elements are enabled. On small screensthe banner always uses the stacked layout.
+
+## Layouts
```
mperrotti, line 246: I think we need more information in this section. It might be helpful to answer the following questions:
- How do I compose a banner into the rest of the page layout?
- When is it ok to wrap the buttons to their own line instead of keeping them on the same line as the title or message?

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+<img
+  width="960"
+  alt="Layouts depending on which elements are enabled"
+  src="https://github.com/primer/design/assets/813754/fe702f23-de88-47a9-b703-b8b262698a50"
+/>
+Text missing
+
+## Best practices
```
mperrotti, line 254: This is really important information. I think we should move this to the top of the page by nesting it under the "Usage" heading.

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+Use banners sparingly and only when necessary. Banners can disrupt the user experience and should only be used when and where relevant.
+
+Keep the message concise and direct. Don't use headings or multiple type sizes in banners, rely on the default paragraph size and use the title when appropiate.
+
+Don't show more than one banner at a time. For more information, see this article on [alert fatigue](https://www.nngroup.com/videos/alert-fatigue-user-interfaces/).
+
 ## Accessibility
```
mperrotti, line 261: Can you confirm that none of the proposed Banner updates affect our existing a11y guidance?

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
 2. To communicate feedback in response to a user action.
 
+For system wide information, consider using an [announcement](../ui-patterns/messaging#message-types).
+
+## Anatomy
+
+<img
+  width="960"
+  alt="Anatomy of a banner, including the message, title, icon, dismiss button, primary, and secondary action"
+  src="https://github.com/primer/design/assets/813754/defa52f8-49bb-4332-aaa1-872a22a563a9"
```
mperrotti, line 35: I'd recommend changing this diagram to only contain a single banner with all parts labeled.

The purple banner in this image has all the parts we'd need.

content/components/banner.mdx
```diff
@@ -8,15 +8,256 @@ tags:
...
+/>
+
+## Options
+
+### Variant
```
mperrotti, line 40: Do you think it would be worth explaining which is the "default" variant?

content/components/banner.mdx
```diff
@@ -8,15 +8,254 @@ tags:
...
 
+import {Box} from '@primer/react'
 import ComponentLayout from '~/src/layouts/component-layout'
 export default ComponentLayout
 import {AccessibilityLink} from '~/src/components/accessibility-link'
 
-Banner can be used in one of two ways:
+<img
+  width="960"
+  alt="Image displaying each variant of banenr, from top: critical, warning, success, info, upsell."
```
langermank, line 18: This is quite nitpicky, but the specific text you've used in all the images is for a message I think we would see in a global Announcement above the navbar. Is there something more generic you could use to not cause confusion?

content/components/banner.mdx
langermank: Agree with Mike 👍

content/components/banner.mdx
```diff
@@ -8,15 +8,254 @@ tags:
...
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="critical banner"
+    src="https://github.com/primer/design/assets/813754/8b135c8e-898d-423c-81c4-9dbb4736a4d5"
+  />
+  <Box as="p" mt="0">
+    <h4>Critical</h4>
+    Use critical banners to signal critical errors, system failures, or necessary destructive actions. Examples: form validation error summary, repo deletion confirmations.
```
langermank, line 56: I would also consider using custom language for each of these variants that align with the severity.

content/components/banner.mdx
```diff
@@ -8,15 +8,254 @@ tags:
...
+>
+  <img
+    width="456"
+    alt="success banner"
+    src="https://github.com/primer/design/assets/813754/fe04a355-d343-4a79-9948-efc64d6de76e"
+  />
+  <Box as="p" mt="0">
+    <h4>Success</h4>
+    Use success banners to indicate task completion or successful actions. For example to confirm saving of settings or successful issue transfer. Only use success messages if the success is not easily apparent in the UI.
+  </Box>
```
langermank, line 93:
```suggestion
    Use success banners to indicate task completion or successful actions. For example, to confirm settings have been saved or a successful issue transfer. Only use success messages if the success is not easily apparent in the UI.
  </Box>
```

content/components/banner.mdx
```diff
@@ -8,15 +8,254 @@ tags:
...
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="banner without title and banner with title"
+    src="https://github.com/primer/design/assets/813754/c5876803-7b5b-486d-b2ec-27fcfdf0c5d8"
+  />
+  <Box as="p" mt="0">
+    <h4>Title</h4>
+    Use a title to consisly present the topic of the banner when the message of the banner is complex and too long to scan.
```
langermank, line 148: Can you add something about title being required but may be visually hidden?

content/components/banner.mdx
```diff
@@ -8,15 +8,254 @@ tags:
...
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="banners with a dismiss button"
+    src="https://github.com/primer/design/assets/813754/e9863c48-cf3a-472b-a066-0ba139a633ca"
+  />
+  <Box as="p" mt="0">
+    <h4>Dismiss button</h4>
+    Banners with an <i>onDismiss</i> property automatically show a dismiss button.
```
langermank, line 168:
```suggestion
    Banners with an `onDismiss` property automatically show a dismiss button.
```

langermank: Maybe we should explain why you would or wouldn't show the dismiss button?

content/components/banner.mdx
```diff
@@ -8,15 +8,254 @@ tags:
...
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="banners with icons, some of which are customized"
+    src="https://github.com/primer/design/assets/813754/dd6d9159-8de7-4f71-b69b-ede15284aeb4"
+  />
+  <Box as="p" mt="0">
+    <h4>Icon</h4>
+    The icon relates to the selected <a href="#variant">variant</a> and can't be disabled. It can only be changed for banners of variant <i>info</i> or <i>upsell</i>.
```
langermank, line 231:
```suggestion
    The icon relates to the selected <a href="#variant">variant</a> and can't be disabled. It can only be changed for banners of variant `info` or `upsell`.
```

content/components/banner.mdx
```diff
@@ -8,15 +8,254 @@ tags:
...
+</Box>
+
+## Breakpoints
+<img
+  width="960"
+  alt="Anatomy of a banner, including the message, title, icon, dismiss button, primary, and secondary action"
+  src="https://github.com/primer/design/assets/813754/b1db8982-43ce-4a95-b25a-407020b53440"
+/>
+
+On large screens the layout of the elements within the banner adjust depending on which elements are enabled. On small screensthe banner always uses the stacked layout.
```
langermank, line 242:
```suggestion
On large screens the layout of the elements within the banner adjust depending on which elements are enabled. On small screens the banner always uses the stacked layout.
```

content/components/banner.mdx
langermank: Maybe this section should be about `Position` to show where Banners might be rendered: inside a Dialog, under the global nav, top of a section etc etc

langermank (commented):
Good start!

content/components/banner.mdx
```diff
@@ -8,15 +8,254 @@ tags:
...
+  width="960"
+  alt="Layouts depending on which elements are enabled"
+  src="https://github.com/primer/design/assets/813754/fe702f23-de88-47a9-b703-b8b262698a50"
+/>
+Text missing
+
+## Best practices
+Use banners sparingly and only when necessary. Banners can disrupt the user experience and should only be used when and where relevant.
+
+Keep the message concise and direct. Don't use headings or multiple type sizes in banners, rely on the default paragraph size and use the title when appropiate.
```
joshblack, line 255:
```suggestion
Keep the message concise and direct. Don't use headings or multiple type sizes in banners, rely on the default paragraph size and use the title when appropriate.
```

joshblack: One question, when it says "Don't use headings" is that in addition to the default heading present in the banner?

joshblack (commented):
Looks great! Just left a comment/question.

content/components/banner.mdx
lukasoppermann: @joshblack is this how we implement it, or is there a separate a11y title which will just use the title the user defines if they do?

content/components/banner.mdx
lukasoppermann: I used italics because for me the backticks don't result in code blocks within the `<Box>` component. Am I missing something?

> Maybe we should explain why you would or wouldn't show the dismiss button?
Yes, great point. Do you already have something in mind here?

content/components/banner.mdx
lukasoppermann: Same as above, how do I get this to work inside of box?

content/components/banner.mdx
lukasoppermann: Yes, maybe a do-don't block would be good here. It means to not add e.g. an `h4` inside of the body of the banner.

content/components/banner.mdx
lukasoppermann: No, I'd love to get this info form @ichelsea

content/components/banner.mdx
lukasoppermann: Yeah, I thought about this as well. Do you have good examples from GitHub in mind, or should we make up random ones?

content/components/banner.mdx
lukasoppermann: I am not sure there should be a default. I think it makes sense to require the user to always set the variant property, so they actively decide for a variant.

content/components/banner.mdx
lukasoppermann: Can you elaborate on this, or add a suggestion? I am not sure I understand what you mean.

content/components/banner.mdx
lukasoppermann: > When is it ok to wrap the buttons to their own line instead of keeping them on the same line as the title or message?

This happens automatically, but still interesting to explain.

> Maybe this section should be about Position to show where Banners might be rendered: inside a Dialog, under the global nav, top of a section etc etc

Yes, this sounds pretty helpful. 👍

content/components/banner.mdx
```diff
@@ -5,18 +5,265 @@ componentId: banner
...
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="banners with a dismiss button"
+    src="https://github.com/primer/primitives/assets/813754/85c94192-eb1a-4e7c-8fc0-b3c5f31ac251"
+  />
+  <Box as="p" mt="0">
+  
+  <h4>Dismiss button</h4>
```
lukasoppermann, line 174:
```suggestion
  <h4>Dismiss button</h4>

```

content/components/banner.mdx
ichelsea: Adding this to my list of things to look at this week!

content/components/banner.mdx
mperrotti: Ah ok, so `variant` is a required prop.

content/components/banner.mdx
langermank: Aw, I thought it would work in `Box`. I guess italics is fine 🤷

content/components/banner.mdx
langermank: If you look through the decision making you logged around close buttons (like maybe in the epic updates?) I think you might find something from accessibility conversations? I don't have the answer here off the top of my head. If its an option to not show it, we should have a reason why we offer that and then explain is 😄

content/components/banner.mdx
langermank: Sorry, I'm referring to the images used for each case. So instead of using the same language in each image for each variant, try and be more specific.

langermank: Just checked the deploy and I think your updates are awesome! 👍

content/components/banner.mdx
langermank: I think what you came up with looks perfect! 🎉

langermank: Only thing I'm noticing is this image feels really tiny and hard to read. What do you think?

header image for banner

lukasoppermann: Yeah, true, I'll resize it.

content/components/banner.mdx
lukasoppermann: @mperrotti  helped me. Apparently you need blank lines inside the box and the content can't be indented to be recognized as markdown.

content/components/banner.mdx
lukasoppermann: Hey, I didn't find the section you mention. It is also not really an a11y concern I think.

I added some more context to the section though.

content/components/banner.mdx
lukasoppermann: I am not sure if this is the right place for a discussion, but I am a bit afraid that if the title is not hidden by default, we will get a lot of duplicates with title and body being the same text.

My only thought to improve this would be to have two props like `aria-title` and `title` where `aria-title` is set to `title` if present. So, using typescript we could require on of the two to be set.

But I am no react dev and this may be an uncommon or anti-pattern.

mperrotti:
Looks good to me. I'd "approve" but waiting until this is out of "draft".

Just ping me when you're officially ready for a review 🙂

content/components/banner.mdx
ichelsea: Okay, this entire a11y section was written up form the work that @lindseywild and @khiga8 did, so it's valid! There are a couple of things that we want to adjust based on feedback from engineering office hours earlier this week that has also been on my to-do list to get to. 

We can create a separate PR for that if you want to get this one merged, or I can add to it before this one is merged.

ichelsea (changes requested):
Thanks for getting this documentation going, @lukasoppermann!

I have a couple fundamental thoughts around our suggestions vs our inline messaging that's coming. Should we write like we don't yet have that component or write like we will have that component soon?

content/components/banner.mdx
```diff
@@ -5,18 +5,293 @@ componentId: banner
...
+  flexDirection={['column', 'column', 'column', 'column', 'row']}
+  sx={{gap: 4}}
+>
+  <img
+    width="456"
+    alt="banners with icons, some of which are customized"
+    src="https://github.com/primer/primitives/assets/813754/9de199af-6488-4ae3-b4a7-14896f5cb719"
+  />
+
+The icon relates to the selected [variant](#variant) and can't be disabled. It can only be changed for banners of variant `info` or `upsell`.
```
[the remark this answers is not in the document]
lukasoppermann: No suggestions, it very much depends on the context.

content/components/banner.mdx
lukasoppermann: Please add it to this PR. We will keep it open until the react components are implemented and shipped, so it will be open for a bit.

content/components/banner.mdx
```diff
@@ -5,18 +5,293 @@ componentId: banner
...
+  <img
+    width="456"
+    alt="trailing actions on a banner"
+    src="https://github.com/primer/primitives/assets/813754/7c57bc61-ee0d-4673-86e9-550b945b117a"
+  />
+  <div>
+
+A maximum of 2 actions can be added to a banner. Whenever possible it is recommended to only add a single action to better direct the users attention.
+
+For banners with a single action, you can choose between adding a primary or secondary action depending on how much attention you want to draw to the action.
```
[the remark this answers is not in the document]
lukasoppermann: I see what you are saying. However, we have a default button and a invisible button here. The default has a lot more visual prominence.

I am happy to update names (please suggest something that works better), but I think it is important to keep a reference to one being more prominent and if used together this one should be the "suggest" action, hence primary in the current naming.

We don't know the context, sometimes a single button action may be very important, sometimes it may be less so, which is why consumer can choose 
the type.

content/components/banner.mdx
```diff
@@ -5,18 +5,293 @@ componentId: banner
...
+  <img
+    width="456"
+    alt="critical banner"
+    src="https://github.com/primer/primitives/assets/813754/9c245a1e-5d59-44bf-8574-aedb06a286b7"
+  />
+  <div>
+
+#### Critical
+
+Use critical banners to signal critical errors, system failures, or necessary destructive actions. Examples: form validation error summary, repo deletion confirmations.
```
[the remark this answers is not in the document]
lukasoppermann: This is not talking about the once it is deleted, but rather before you delete.

However, I think this was somewhat recently removed. Is there another good example?

content/components/banner.mdx
```diff
@@ -5,18 +5,281 @@ componentId: banner
...
+
+<Box
+  mb={3}
+  display="flex"
+  alignItems="flex-start"
+  flexDirection={['column', 'column', 'column', 'column', 'row']}
+  sx={{gap: 4}}
+>
+
+Banners with an `onDismiss` property automatically show a dismiss button. This should be used for banners that inform about something that the user can not solve. For example issues with the organization they are part of or informational banners. When it makes sense, the choice to dismiss the banner should be stored and the banner should not be shown again.
```
lukasoppermann, line 173:
```suggestion
Banners with an `onDismiss` property automatically show a dismiss button. This should be used for banners that inform about something that the user can not solve. For example issues with the organization they are part of or informational banners. When it makes sense, the choice to dismiss the banner should be stored and the banner should not be shown again.

Use the following as a guideline:
- If you're on a page in which you can do other tasks, and the banner is not blocking / does not have to be addressed, then you should be able to have a close button.
- If accidental closure of the banner would leave the user confused or missing something critical, it should not have a close button.
```

content/components/banner.mdx
```diff
@@ -5,18 +5,285 @@ componentId: banner
...
+
+Use the following as a guideline:
+- If you're on a page in which you can do other tasks, and the banner is not blocking / does not have to be addressed, then you should be able to have a close button.
+- If accidental closure of the banner would leave the user confused or missing something critical, it should not have a close button.
```
lukasoppermann, line 177:
```suggestion
* If you're on a page in which you can do other tasks, and the banner is not blocking / does not have to be addressed, then you should be able to have a close button.
* If accidental closure of the banner would leave the user confused or missing something critical, it should not have a close button.
```

joshblack (approved):

lukasoppermann:
@ichelsea I merged now as the banner already shipped, please add your improvements via a separate PR. 🙏

PR #751, opened 2024-04-03
Use the `docsId` field to group multiple versions of a React component

camertron:
This PR allows a `docsId` field to group multiple versions of a React component together into a single page. The `docsId` concept is motivated by the need to generate a single page for the `Tooltip` component, which has two versions named `tooltip` and `tooltip_v2`.


broccolinisoup (approved):
🎉

camertron:
Note that the checks for this PR won't pass until https://github.com/primer/react/pull/4458 is merged and released.

PR #752, opened 2024-04-03
Read import path from component json

broccolinisoup:
Having a new next entry point made the current auto generated component import path invalid for next components such as tooltip. Discussed this with @camertron and instead of generating the import path based on the component status, we think adding another `importPath` as a field on the component.docs.json and read the import path directly from field is less prune to invalid path generation. This PR alters how we form the import path.

Associated Primer React PR: https://github.com/primer/react/pull/4461

### DIFF
--- a/gatsby-node.esm.js
+++ b/gatsby-node.esm.js
@@ -378,7 +378,6 @@ exports.createPages = async ({actions, graphql}) => {
           name
           status
           a11yReviewed
-          importPath
           stories {
             id
             code
@@ -621,7 +620,6 @@ exports.onPostBuild = async ({graphql}) => {
           name
           status
           a11yReviewed
-          importPath
           stories {
             id
             code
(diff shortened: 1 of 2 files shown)


gatsby-node.esm.js
```diff
@@ -378,6 +378,7 @@ exports.createPages = async ({actions, graphql}) => {
           name
           status
           a11yReviewed
+          importPath
```
broccolinisoup, line 381: @camertron I am not super sure about the graphql schema updates. Feel free to edit it on the PR or suggest changes if they are not correct.

camertron, line 381: We only need to add `importPath` here if we're using it to build pages, which it appears we're not. We should be able to remove this line and the other one in this file 👍 Not a blocker tho!

camertron (approved):
Nice!!

gatsby-node.esm.js
broccolinisoup: Hmm, okay. Thanks! I'll remove it now.

broccolinisoup, line 381:
```suggestion
```

camertron (approved):

PR #765, opened 2024-04-23
Band aid header

dipree:
Making sure every page uses a layout with the much needed header instead of using anything from the template.

Re-did:
- layout with toc but not a component
- simple layout w/o toc (special treatment for source)
- hero layout


joshblack (commented):
Just one question on the layouts, is it possible for us to use a default layout for these so we don't have to wrap each page? e.g. https://www.npmjs.com/package/gatsby-plugin-mdx/v/1.0.13#default-layouts or do we need to add explicit layouts for each page?

cc @camertron curious what you'd recommend 👀

dipree:
@joshblack I tried that first but it's resulting in a number of errors. It's also worth noting that all the component pages already use a layout, but not "default" pages while some index pages had the index layout, some had not.

PR #767, opened 2024-04-24
feat: add analytics to gatsby-browser config

joshblack:
This is a follow up to: https://github.com/primer/design/pull/765 to see if we can add analytics using `gatsby-browser.js` and the `wrapPageElement` helper there. This helps to prevent having to implement custom layouts and export them in content files.


dipree (approved):
Thanks @joshblack I could validate events are being sent.

PR #782, opened 2024-05-02
Heading component: Publish a11y docs on Primer.style

auareyou:
Added modifications per: https://github.com/github/primer/issues/3245


content/components/heading.mdx
```diff
@@ -28,12 +28,46 @@ Headings are used to structure the content on the page, with each heading repres
...
 
 Headings allow assistive technology users to quickly navigate around a page. Navigation to text that is not meant to be a heading can be a confusing experience. [Learn more about best heading practices (WAI Headings)](https://www.w3.org/WAI/tutorials/page-structure/headings/).
 
+## Built-in accessibility features
+
+The heading component exposes both the heading role and heading level.
+
+The heading component provides styles with appropriate text sizes to provide a visual hierarchy when the appropriate level is set.
+
+The default heading text has sufficient color contrast against a white background.
```
mperrotti, line 51: We can probably remove this because this is true of all text, not just heading text. I don't mind keeping it though.

mperrotti (approved):

PR #796, opened 2024-06-01
include Typography utility class names in documentation for easier reference

ktravers:
### Context

Whenever I look up typography utility classes, I end up having to inspect the markup in the embedded Storybook to get the actual class name. It would be much easier if the class names were written out in the documentation instead, so they could be copy/pasted directly from the page, without having to take the extra step of inspecting the page source.

### Summary

This PR adds the class names to the documentation. I tried to maintain consistency with how class names are already referenced elsewhere in the doc (ex. header class names).

### Related 

https://github.com/primer/css/pull/2633


langermank (approved):
Thank you! 😄

PR #803, opened 2024-06-20
Add further clarification on padding in our layout documentation

maximedegreve:
I've observed still quite a bit of confusion around the padding in our layouts. This PR aims to clarify our expectations when it comes to padding in our layouts.


langermank (approved):
Nice improvement!

PR #808, opened 2024-06-25
Add mention of dartsass-rails

mathias:
I think this is correct but I haven't verified with local setup


keithamus (approved):
Seems reasonable enough to add this.

PR #809, opened 2024-06-25
Update layout in banner.mdx

joshblack:
Noticed a layout issue in Banner that was being caused by the flex layout plus list in the content. In this case, it was causing the list to be formatting oddly 🤔 

I wrapped this section in a `div` for what I think would be the correct layout but feel free to propose a different chase if this is incorrect!


joshblack:
Closing since: https://github.com/primer/design/pull/809 was merged!

PR #813, opened 2024-07-01
Fix broken layout in banner.mdx

patrickhlauke:
Adds wrapping `<div>` to fix the broken layout for the https://primer.style/components/banner#dismissable section in the documentation.

screenshot of the currently broken documentation section, showing content awkwardly split over four columns when it should only be two columns

### DIFF
--- a/content/components/banner.mdx
+++ b/content/components/banner.mdx
@@ -176,8 +176,10 @@ Banners with an `onDismiss` property automatically show a dismiss button. This s
 
 Use the following as a guideline:
 
-* If you're on a page in which you can do other tasks, and the banner is not blocking / does not have to be addressed, then you should be able to have a close button.
-* If accidental closure of the banner would leave the user confused or missing something critical, it should not have a close button.
+<ul>
+  <li>If you're on a page in which you can do other tasks, and the banner is not blocking / does not have to be addressed, then you should be able to have a close button.</li>
+  <li>If accidental closure of the banner would leave the user confused or missing something critical, it should not have a close button.</li>
+</ul>
 
   </div>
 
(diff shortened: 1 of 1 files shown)


content/components/banner.mdx
```diff
@@ -170,12 +170,16 @@ Use upsell banners to inform the users about a feature that can be enabled by up
   sx={{gap: 4}}
 >
 
+  <div>
+
 Banners with an `onDismiss` property automatically show a dismiss button. This should be used for banners that inform about something that the user can not solve. For example issues with the organization they are part of or informational banners. When it makes sense, the choice to dismiss the banner should be stored and the banner should not be shown again.
 
 Use the following as a guideline:
```
patrickhlauke, line 177:
```suggestion
Use the following as a guideline:

```

keithamus (approved):

patrickhlauke:
I'm unsure why the preview/build step is failing here (as I only took inspiration from the following section in the doc that used the approach of wrapping things in a `<div>` - if anybody knows/can fix this, I'd be grateful /cc @keithamus

patrickhlauke:
Does it need to have the unordered list written out fully as HTML as well, to work around this?

content/components/banner.mdx
```diff
@@ -170,12 +170,17 @@ Use upsell banners to inform the users about a feature that can be enabled by up
...
+
 * If you're on a page in which you can do other tasks, and the banner is not blocking / does not have to be addressed, then you should be able to have a close button.
```
patrickhlauke, line 179:
```suggestion
<ul>
  <li>If you're on a page in which you can do other tasks, and the banner is not blocking / does not have to be addressed, then you should be able to have a close button.</li>
```

content/components/banner.mdx
```diff
@@ -170,12 +170,17 @@ Use upsell banners to inform the users about a feature that can be enabled by up
...
 * If accidental closure of the banner would leave the user confused or missing something critical, it should not have a close button.
```
patrickhlauke, line 180:
```suggestion
  <li>If accidental closure of the banner would leave the user confused or missing something critical, it should not have a close button.</li>
</ul>
```

keithamus (approved):
Strange that it needed that... but sure, still LGTM :+1:

patrickhlauke:
Some shortcoming in the markdown converter it seems, but yes confirming that visually the preview now looks good https://primer-8e64b013b2-26441320.drafts.github.io/components/banner#dismissable

Merging (I'm assuming it's ok for me to merge here once it's been approved - if not, please let me know if there's a more official process to follow)

PR #821, opened 2024-07-12
Add link to full size flowchart in notification-messaging

dfosco:
Added link to view the flowchart in full size right before "View image description":

image

image

I don't think our Primer docs doesn't have a lightbox component to expand images, so linking to the full size image is the best way to view it in detail.

### DIFF
--- a/content/ui-patterns/notification-messaging.mdx
+++ b/content/ui-patterns/notification-messaging.mdx
@@ -217,7 +217,9 @@ Picking a notification message for an experience can be tricky. Reference this f
   alt=""
   src="https://github.com/user-attachments/assets/b8635f19-629c-42d6-9102-0014257288e5"
 />
+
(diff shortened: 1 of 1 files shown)


content/ui-patterns/notification-messaging.mdx
```diff
@@ -217,7 +217,7 @@ Picking a notification message for an experience can be tricky. Reference this f
   alt=""
   src="https://github.com/user-attachments/assets/b8635f19-629c-42d6-9102-0014257288e5"
 />
-
+[View flowchart in full size](https://github.com/user-attachments/assets/b8635f19-629c-42d6-9102-0014257288e5)
```
langermank, line 220:
```suggestion

[View flowchart in full size](https://github.com/user-attachments/assets/b8635f19-629c-42d6-9102-0014257288e5)

```

I think this markup is funky if there's no spaces (mdx being picky)

langermank (approved):

PR #824, opened 2024-07-17
Remove expectation of breaking changes from Alpha

siddharthkp:
```diff > ## Alpha

- The component is ready for preliminary usage, with breaking changes expected.
+ The component is ready for preliminary usage
```

Removed the part about breaking changes expected because this does not match our processes anymore. We only mark a component as alpha when the API is stable. (We use experimental components for unstable APIs)

> [!IMPORTANT]  
> Need to confirm if this is true for PVC as well.

### DIFF
--- a/content/guides/component-lifecycle.mdx
+++ b/content/guides/component-lifecycle.mdx
@@ -13,7 +13,7 @@ The component exists as a proof-of-concept or is work in progress.
 
 ## Alpha
 
-The component is ready for preliminary usage, with breaking changes expected.
+The component is ready for preliminary usage
 
 - The component does not have external dependencies. This could be external libraries or dependencies on libraries in an application (like GitHub.com)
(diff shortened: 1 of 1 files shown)


joshblack (approved):

content/guides/component-lifecycle.mdx
```diff
@@ -13,7 +13,7 @@ The component exists as a proof-of-concept or is work in progress.
 
 ## Alpha
 
-The component is ready for preliminary usage, with breaking changes expected.
+The component is ready for preliminary usage
```
jonrohan, line 16:
```suggestion
The component is ready for preliminary usage.
```

jonrohan (approved):
This is typical for PVC as well

content/guides/component-lifecycle.mdx
siddharthkp: thanks!

PR #839, opened 2024-08-21
Fix flex utilities link in Grid

mjimenez98:
### Problem

Flex utilities link in https://primer.style/foundations/css-utilities/grid returns a 404. It points to https://primer.style/css-utilities/flexbox/

Screenshot 2024-08-21 at 16 05 05

### Solution

Point to the correct link: https://primer.style/foundations/css-utilities/flexbox/


camertron (approved):
Thank you ❤️

PR #844, opened 2024-08-26
Add note on icons in UnderlinePanels

TylerJDev:
Closes https://github.com/github/primer/issues/3579

Clarifies that icons in `UnderlinePanels.Tab` are purely decorative.

### DIFF
--- a/content/components/underline-panels.mdx
+++ b/content/components/underline-panels.mdx
@@ -51,6 +51,8 @@ Underline panels are made of the following ARIA roles:
 
 The active tab is conveyed to assistive technologies using the `aria-selected` attribute. Each tab is associated with it's corresponding panel using the `aria-controls` attribute.
 
+Icons in [UnderlinePanels.Tab](https://primer.style/components/underline-panels/react/draft#underlinepanelstab) are purely decorative and are hidden from screen readers by default. If the icon does convey any informational meaning, the adjacent text inside of the tab should provide all the necessary information without relying on the icon itself.
+
 </div>
 
 <img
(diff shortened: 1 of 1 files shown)


content/components/underline-panels.mdx
```diff
@@ -51,6 +51,8 @@ Underline panels are made of the following ARIA roles:
 
 The active tab is conveyed to assistive technologies using the `aria-selected` attribute. Each tab is associated with it's corresponding panel using the `aria-controls` attribute.
 
+Icons in [UnderlinePanels.Tab](https://primer.style/components/underline-panels/react/draft#underlinepanelstab) are purely decorative and are hidden from screen readers by default. If the icon does convey any informational meaning, the adjacent text inside of the tab should provide all the necessary information without relying on the icon itself.
```
patrickhlauke, line 54:
```suggestion
Icons in [UnderlinePanels.Tab](https://primer.style/components/underline-panels/react/draft#underlinepanelstab) are purely decorative and are hidden from screen readers by default. If the icon does convey any informational meaning, the adjacent text inside of the tab must provide all the necessary information without relying on the icon itself.
```

patrickhlauke (approved):
Thanks @TylerJDev ... looks good, though I'd suggest changing the "should" to "must" (unless that's a bit too strict/lacks nuance?)

content/components/underline-panels.mdx
TylerJDev: This is a good suggestion!

lukasoppermann (approved):

PR #850, opened 2024-08-29
Add note on `Timeline.Break`

TylerJDev:
Closes https://github.com/github/primer/issues/3355

Clarifies that `Timeline.Break` should be used for decorative purposes only.


camertron (approved):
Awesome! 🎉

PR #851, opened 2024-08-29
Update tools.mdx

janmaarten-a11y:
Updating links and references from the old A11y Design toolkit to the new Annotation Toolkit


content/guides/accessibility/tools.mdx
```diff
@@ -11,11 +11,12 @@ When testing contrast be wary of tools that use color pickers. The picked color
 
 ## Figma
 
-### [Accessibility sticker sheet](https://www.figma.com/file/GQEYQhQ5s19Wsp6p2wMcU0/A11y-Design-resources?node-id=1-2&t=SsZrSUTm6mbV06jp-11)
```
janmaarten-a11y, line 14: Removed the direct link to avoid a lot of external file access requests, especially since the library is already active for folks.

ericwbailey (approved):

PR #857, opened 2024-09-09
Add note on `NavList.LeadingVisual`

TylerJDev:
Closes https://github.com/github/primer/issues/3427

Adds note on `LeadingVisual` in `NavList` being decorative only.


francinelucca (approved):
👍🏽

PR #860, opened 2024-09-13
docs(PageHeader): add a11y guidance for role and aria-label

francinelucca:
Adds accessibility guidance for usage of `PageHeader`'s `role` and `aria-label` props when a landmark hasn't been previously specified. Relates to https://github.com/github/primer/issues/3466

image

See deployment: https://primer-6a854b4d02-26441320.drafts.github.io/components/page-header#landmark-usage

### DIFF
--- a/content/components/page-header.mdx
+++ b/content/components/page-header.mdx
@@ -304,7 +304,7 @@ The title bar can include a static page title, actions, and leading and trailing
 Pageheader allows usage of different HTML landmarks to create a more accessible UI.
 The usage of `header` and `nav` regions are allowed within PageHeader, both with different use cases.
 
-If no landmarks are used within the `PageHeader`, consider using the `role` prop to indicate one and provide it with an accessible name (perhaps the same as `PageHeader.Title`) via `aria-label`.
+If the `PageHeader` is not contained within or contains a landmark, consider using the `role` prop to indicate one and provide it with an accessible name (perhaps the same as `PageHeader.Title`) via `aria-label`.
 
 #### PageHeader with `header` landmark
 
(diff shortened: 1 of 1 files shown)


TylerJDev (approved):

francinelucca:
Can I get a check on this from @primer/accessibility-design? Thanks! tagging FR @ichelsea

content/components/page-header.mdx
```diff
@@ -304,6 +304,8 @@ The title bar can include a static page title, actions, and leading and trailing
 Pageheader allows usage of different HTML landmarks to create a more accessible UI.
 The usage of `header` and `nav` regions are allowed within PageHeader, both with different use cases.
 
+If the `PageHeader` is not contained within or contains a landmark, consider using the `role` prop to indicate one and provide it with an accessible name (perhaps the same as `PageHeader.Title`) via `aria-label`.
```
ericwbailey, line 307:
```suggestion
If the `PageHeader` is not contained within a landmark, consider using the `role` prop to indicate one and provide it with an accessible name (perhaps the same as `PageHeader.Title`) via `aria-label`.
```

ericwbailey, line 307: Do you think it'd be worth providing a high-level example of the desired code here? Saying this because this is an area that engineers are often unfamiliar with, so presenting a good default to copy from might be helpful.

francinelucca: yeah, good idea!

francinelucca: I agree this wording is kind of weird, I'm trying to guide users to prevent a "Banner landmark cannot be nested within another landmark." error , which means there shouldn't be a landmark up in the chain or within the chain, but maybe we do not need to get that deep into it? 🤔

ericwbailey (approved):

PR #863, opened 2024-09-18
Add note on visually hidden labels

TylerJDev:
Adds a note on visually hidden labels, that there must be a visual label defined if the label is visually hidden.

Context: https://github.com/github/primer/issues/3441

### DIFF
--- a/content/components/textarea.mdx
+++ b/content/components/textarea.mdx
@@ -49,7 +49,7 @@ For more information about the "valid" and "invalid" states, see the [validation
   src="https://user-images.githubusercontent.com/6905903/274239156-064abae9-6d16-4c1a-b673-ec8f96429ed9.png"
 />
 
-**Label (required):** The input's title. It should be as concise as possible and convey the purpose of the input. The label may be visually hidden in rare cases, but a label must be defined for assistive technologies such as a screen reader.
+**Label (required):** The input's title. It should be as concise as possible and convey the purpose of the input. The label may be visually hidden in rare cases, but a label must be defined for assistive technologies such as a screen reader. If the label is visually hidden, ensure there is an element nearby that serves as a visual label for the input. This guarantees that all users, including those not using assistive technologies, have access to a label.
 
 **Required indicator:** Indicates that a value is required. Must be shown for any required field, even if all fields in the form are required.
 


TylerJDev:
@primer/accessibility-design - Would love a review on this PR, as I believe we can add this note to other form elements that describe usage of "visually hidden" labels.

content/components/textarea.mdx
```diff
@@ -49,7 +49,9 @@ For more information about the "valid" and "invalid" states, see the [validation
   src="https://user-images.githubusercontent.com/6905903/274239156-064abae9-6d16-4c1a-b673-ec8f96429ed9.png"
 />
 
-**Label (required):** The input's title. It should be as concise as possible and convey the purpose of the input. The label may be visually hidden in rare cases, but a label must be defined for assistive technologies such as a screen reader.
+**Label (required):** The input's title. It should be as concise as possible and convey the purpose of the input. The label may be visually hidden in rare cases, but a label must be defined for assistive technologies such as a screen reader. If the label is visually hidden, ensure there is an element nearby that serves as a visual label for the input. This guarantees that all users, including those not using assistive technologies, have access to a label.
```
ericwbailey, line 52: What is an example of a rare case? I worry that without defining that everyone will think their use case is the exception to the rule.

IanPouncey:
I agree with @ericwbailey, more examples would be beneficial.

The most common example is a single field search form, with a magnifying glass icon as the content of the submit button. In this situation, the icon is acting as both the content of the button and the label for the input field.

If an icon is used, it must be unambiguous as to its purpose. From a users point of view, they should be able to say the name of the icon and have voice control software associate the name with the control. To continue the above example, the magnifying glass indicates "search". There are very few other icons that have this same association as they relate to input fields, a few more that could apply to button elements (such as "+" for "add", "plus", or "increase"), so it should be exceedingly rare for an icon to be the only label, whether it is part of the `<label>` element or not.

If another piece of text is used, the accessible name for the `<input>` **must** include, as an unbroken sub-string or as the entire accessible name, the visible text.

TylerJDev:
Here are a few examples I could think of right now:

* On an issues/PR page, you can edit the title which brings up an input of the existing title as a value. I'm not entirely sure how _accessible_ this is, the context is there when you activate the "edit" button, but once activated, that context is lost.
* GitHub's search query utilizes an input without a visual text label, but contains an icon (magnifying glass). This is mostly the same as @IanPouncey's example.
  * I'd say this is the most common usage of the prop. Inputs for searching files/branches consist of the same pattern, usually accompanied by a `placeholder` and icon.
* We recommend this pattern in our docs [for nested form controls specifically](https://primer.style/ui-patterns/forms/overview#nested-form-controls). I'm not too sure how often this pattern is utilized but I'm curious if this approach is accessible, the context seems to be there initially. 

[We use this prop a lot](https://musical-adventure-wlr3n3k.pages.github.io/?name=formcontrol&attribute=%22visuallyHidden%22), so if we want to stop suggesting it entirely we could definitely do so. Thanks too for the input 🙇

PR #870, opened 2024-10-04
docs: add rails version to Stack component, remove ImageCrop docs

francinelucca:
Stack component has been introduced to Primer view components, adding it so it here so it can be reflected in our docs. 

Also deletes the ImageCrop component docs and nav item since this component was removed in https://github.com/primer/view_components/pull/3158


camertron (approved):
🎉

francinelucca:
Will merge when https://github.com/primer/view_components/pull/3085 releases

camertron (approved):

PR #882, opened 2024-10-25
fix: allow status table to render with partial information

francinelucca:
Allow status data to render even when partial information is available (i.e. one of the sources fails to load). 
See deployment: https://primer-fe30d4f93e-26441320.drafts.github.io/guides/status
Currently it will fail to render and get stuck loading:
image

With fix:
image


TylerJDev (approved):
LGTM ✨ I think loading in partial information is better than none at all, as the table itself provides value.

PR #883, opened 2024-10-25
Fix status page

TylerJDev:
Swaps out https://primer.github.io/react/components.json with data we get initially [via unpkg & graphql](https://github.com/primer/design/blob/f575f848ed5f8be315f85013dda8eb9dd6c8c04b/gatsby-node.esm.js#L252-L254). This fixes the issue where the status page doesn't load the component table.

It seems that `https://primer.github.io/react/components.json` was removed, and the [JSON provided by unpkg](https://unpkg.com/@primer/react@37.0.1/generated/components.json) is slightly different, so I modified the function slightly to account for that.

This differs from https://github.com/primer/design/pull/879, as we're utilizing the graphql data that we're already pulling, and aligns more with how we handle rails components on the status page.

### DIFF
--- a/src/components/status.jsx
+++ b/src/components/status.jsx
@@ -224,12 +224,12 @@ async function getComponents(railsActions, railsData, reactData) {
             const current = getStatusIndex(status)
             const existing = getStatusIndex(components[name].status)
 
-            if (current < existing) {
-              components[name] = {a11yReviewed, id, name, status, path: url}
+            if (current > existing) {
+              return;
             }
-          } else {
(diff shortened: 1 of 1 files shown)


TylerJDev:
If we go this route, we may need to clean up or filter out components without pages, (e.g. [`KeybindingHint`](https://primer.style/react/storybook/?path=/story/experimental-components-keybindinghint--default)), as we pull directly from `*docs.json` in `primer/react`.

camertron (approved):

PR #885, opened 2024-10-28
Adjust docs for loading state

talum:
👋🏼 I've read this sentence a few times and I think it's a little confusing. I think it should be "should get" but I've been reading it as "gets."


TylerJDev (approved):
LGTM, thank you too! ✨

PR #889, opened 2024-11-05
TreeView: Add note on truncation

TylerJDev:
Add quick note on truncation in `TreeView`. Context: https://github.com/primer/react/pull/5122

### DIFF
--- a/content/components/tree-view.mdx
+++ b/content/components/tree-view.mdx
@@ -43,7 +43,7 @@ Node labels should be a succinct title for whatever entity the node represents. 
 
 By default, labels are truncated to a single line. However, node labels may be wrapped to multiple lines in cases where the entire label must always be visible.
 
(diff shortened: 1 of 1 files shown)


keithamus (approved):

content/components/tree-view.mdx
```diff
@@ -43,7 +43,7 @@ Node labels should be a succinct title for whatever entity the node represents.
 
 By default, labels are truncated to a single line. However, node labels may be wrapped to multiple lines in cases where the entire label must always be visible.
 
-When a node's text is truncated, it should still be accessible to users via a browser-native tooltip.
+When a node's text is truncated, it should still be accessible to users in some fashion. Truncation is only accessibly acceptable if there is some way for the user to see the non-truncated text.
```
patrickhlauke, line 46:
```suggestion
When a node's text is truncated, it should still be accessible to users in some fashion. Truncation is only accessibly acceptable if there is some way for users - including those navigating with a keyboard - to see the non-truncated text.
```

patrickhlauke (approved):
Made a suggested addition, but otherwise LGTM

PR #903, opened 2024-12-16
Add link to primerize docs for primer/react

camertron:
Trying to link to these docs in as many places as possible. Also looking to add it to the new docs (internal-only link): https://github.com/github/collaboration-workflows-flex/issues/1369


lesliecdubs (approved):
LGTM

PR #913, opened 2025-01-22
Update critical banner dismiss section

lukasoppermann:
Update critical banner dismiss section

### DIFF
--- a/content/components/banner.mdx
+++ b/content/components/banner.mdx
@@ -200,7 +200,7 @@ Use the following as a guideline:
 >
   <div>
 
-  You should try to avoid dismissable critical banners. They are mostly used for errors which need to be resolved. As a result, the user resolving the error to get rid of the critical banner. If used to warn about destructive actions, the banner should also not be dismissable.
+  Critical banners communicate an error which the user must resolve in order for the banner to disappear. In certain cases, the banner may be dismissible when notifying the user of a failed action. If used to warn about destructive actions, the banner should also not be dismissible.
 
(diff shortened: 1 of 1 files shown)


content/components/banner.mdx
```diff
@@ -200,7 +200,7 @@ Use the following as a guideline:
 >
   <div>
 
-  Critical banners can't be dismissable. They are mostly used for errors which need to be resolved. As a result you can only get rid of a critical banner by resolving the error. If used to warn about   destructive actions, the banner should also not be dismissable.
+  You should try to avoid dismissable critical banners. They are mostly used for errors which need to be resolved. As a result, the user resolving the error to get rid of the critical banner. If used to warn about destructive actions, the banner should also not be dismissable.
```
francinelucca, line 203: > "As a result, the user resolving the error to get rid of the critical banner." 

I think this sentence reads a little weird, what is it trying to say?

francinelucca (changes requested):

content/components/banner.mdx
joshblack, line 203:
```suggestion
  Critical banners communicate an error which the user must resolve in order for the banner to disappear. In certain cases, the banner may be dismissible when notifying the user of a failed action. If used to warn about destructive actions, the banner should also not be dismissible.
```

Took a stab at this, curious what you think of the wording 🤔

joshblack (approved):

francinelucca (approved):
✨

PR #927, opened 2025-02-11
Remove "Rails implementation only" copy from IconButton tooltip documentation

ktravers:
While working on https://github.com/github/github/pull/361769, I noticed that the tooltip functionality described as "Rails only" in the IconButton docs is also in place for the React implementation.

This PR updates the docs accordingly.


camertron (approved):
Cool, thank you!

PR #928, opened 2025-02-13
update React form component urls

ktravers:
I noticed that the component links in the React Forms doc were broken while reviewing https://github.com/github/github/pull/362188.

The url scheme seems to have changed since these docs were last updated. Looks like it now follows the pattern `https://primer.style/components/:component-name/` for interface guidelines and `https://primer.style/components/:component-name/react/latest` for React implementation.

**Questions for reviewers:**
* Are there any url helpers available? Hard-coding the urls seems brittle and could break again if/when the url scheme changes.
* I've seen "dead link" linters in other repos that fail if any urls in the PR return a >4XX response ([example](https://github.com/tcort/markdown-link-check)). Does something like that seem useful for this repo?

### DIFF
--- a/content/ui-patterns/forms/react.mdx
+++ b/content/ui-patterns/forms/react.mdx
@@ -26,9 +26,9 @@ export default FormsLayout
       one or more values.
     </p>
     <Box display="flex" sx={{gap: 3}}>
-      <a href="../components/autocomplete">Interface guidelines</a>
+      <a href="https://primer.style/components/autocomplete">Interface guidelines</a>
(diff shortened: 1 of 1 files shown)


TylerJDev (approved):

TylerJDev:
> Are there any url helpers available? Hard-coding the urls seems brittle and could break again if/when the url scheme changes.

I agree. I believe there was discussion around this recently for our new docs. I'm forgetting what came of it, but @joshblack might know! 

Thanks for the PR too ✨

